PR #1, opened 2016-02-01
Well Known Extended Purpose Codes for Deterministic Wallets

bgok:
Pavol-

This file is related to BIP proposal that I am working on (https://github.com/keepkey/bips/blob/master/bip-xxxx.md). Are you interested in signing on as a co-author and helping marshal this proposal through the BIP process? Since this extends BIP43 and you are one of the original authors, I think it would be appropriate. Also, notice that this proposal makes the SLIP repository the main source for proposals about HD wallet hierarchy standards. Are you interested in taking on the role of sole maintainer or do you want to share the role?

It would be really cool if the SLIP repository were moved to another organization and renamed, but we can discuss this more in the future.

Ken


prusnak:
Could you please rebase this into a single commit? Thanks!

bgok:
Nevermind.

PR #6, opened 2016-03-01
Auroracoin registration.

BioMike:
I'm member of the Auroracoin dev team. I would like to register Auroracoin, because we are developing support for this coin in coinomi-android.

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -85,6 +85,7 @@ index | hexa       | coin
    82 | 0x80000052 | Quarkcoin
    83 | 0x80000053 | Terracoin
    84 | 0x80000054 | Gridcoin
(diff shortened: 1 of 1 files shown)


prusnak:
Ah, is this scam still alive?

BioMike:
Pardon? The coin was taken over by a new team last year, the original owner destroyed the remaining of the premine. An Icelandic foundation was created and the dev team now consists of multiple developers from various countries (including Iceland). We've released a new wallet (and new one with a big update coming soon), there is an AUR-ISK exchange and on the background (within Iceland) there are things being developed. Some people from the dev team are even picking up non-coin projects that were abandoned by their original authors (for whatever reason), but are still popular.

prusnak:
Did not know about burning of the premine. Thanks for the info.

PR #7, opened 2016-03-06
Add IXCoin BIP44

ahmedbodi:


prusnak:
Which wallet does implement IXCoin?

ahmedbodi:
I'm in the process of implementing it into coinomi

PR #10, opened 2016-03-21
Add Monero to SLIP-0044

fluffypony:
Chosen due to 0x80000080 almost being a palindrome, which seemed important at the time;)


prusnak:
89 in hexa is not 0x80. Please use the same value in dec and hex.

fluffypony:
Sorry about that - done:)

fanatid:
@fluffypony `0x80000080` is a byte palindrome, in hex it will be `0x80000008`, so you're little late :(

fluffypony:
@fanatid yep we missed out, which is why we figured ```0x80000080``` is an almost-palindrome if you just drop the x from the hex representation;)

PR #11, opened 2016-04-14
Update slip-0044.md

BigUps:
Add coin type BigUp as value 100 (hex 64)


prusnak:
Which bip44 wallet implements this coin?

BigUps:
This is for a Coinami wallet that we are working on

BigUps:
This is for a Coinami wallet that we are working on

PR #28, opened 2016-08-26
Add Evotion

vetal20600:


prusnak:
Why skip 97?

vetal20600:
I didn't know I can't skip. I just like 98 better :)

dangershony:
I'll be adding another coin soon I can use 97

PR #41, opened 2016-09-23
Handle both token types (Factoids + ECs) in Factom

carryforward:
I thought the wallet manufacturer would prefer to use the different account indexes for the different coin types, but I was wrong.  A better solution for the wallet provider is to treat the two different tokens as different coin types.

Factom has two internal antispam tokens.  The transferable Factoids are converted to non-transferable Entry Credits, which are used to publish data into the blockchain.  They also use different addressing schemes and different pubkey constructions.

https://github.com/FactomProject/FactomDocs/blob/master/factomDataStructureDetails.md#factoid-transaction


carryforward:
I updated the pull request to link to BIP44 test vectors.

PR #53, opened 2016-11-07
[updates for rootstock & rootstock testnet definitions]

chrisforrester:


chrisforrester:
for rootstock.

the reason testnet is not "1" is because this format doesn't work with how hd node structures should be programmed in general.

if we want to have different wallets running simultaneously, there are security issues with publishing the neutered node at the start of any particular chain.

the goal is to isolate these hd node chains, such that a leak of a private key on one of the chains won't for example mean that the entire wallet is compromised, only the keys that are associated with that one chain.

some may argue that testnets aren't that important in this regard, but I disagree.

slush0:
What the constant 37173 means? Can we use just continuous numbering to avoid confusion?

Also, please remove arithmetics from the table, it will be easier to detect collisions in  future PRs.

chrisforrester:
it's just a constant that's a high enough value to not collide with the
primary slip standards. imo a standard arithmetic addition is the most
useful for these.

so ethereum testnet = ethereum + (testnet constant), zcash testnet = zcash
+ (testnet constant).

prusnak:
If you want to use arithmetic, please use some sane value. 

My suggestion is to use `0x80000089` for RSK and `0x90000089` for RSK testnet.

chrisforrester:
one arbitrary value (2415919104 as you say is "sane" because you like hex,
but ofc has nothing to do with actual code since we're using integers, not
hex values natively) has no more precedence than another :)

if your idea of what that arbitrary value should be conflicts, that's not
really on me.

We're currently using RSK testnet in our released, production builds (as of
jaxx 1.1.7) as the value in this PR, and as per the rules, this should be
enough.

prusnak:
Then let's not include the RSK testnet value in the SLIP and you are free to use whichever value you want to use.

chrisforrester:
so because I didn't use a value you didn't tell me about because you're not
working on the features, you're not going to include it? please be a bit
more professional.

y'all are so controlling over things you have no vision over, it's insane.
I'm sure there are some people that have the power to approve things here
that aren't discriminating based on pure spite and/or pride.


RSK Testnet is 37310. There are plenty of instances in this list where the
values are not continuous.

prusnak:
Added in https://github.com/satoshilabs/slips/commit/20e1120aea9a1310601cb61ea7841f1a26c18f63

slush0:
@chrisforrester For next time, I prefer to stay focused on topic and avoid ad hominem.

PR #47, opened 2016-10-18
Update slip-0044.md

insomp:
Just want to update that ZCoin(https://zcoin.tech) will use 88

PR #55, opened 2016-11-16
add Giftblock to SLIP-0044

martintreurnicht:


prusnak:
Please add the entry to the correct place.

martintreurnicht:
@prusnak I moved it to be just after 137

slush0:
What's that blank record for 135?

prusnak:
135 was (and is still) unused.

slush0:
There's many gaps in numbering. For example 40 is unused, too. Still, there's no blank record. So should we remove that blank record or fill gaps to indicate there're unused places? In that case we should fill the table up to Rootstock testnet ;). 

Sorry for being offtopic here.

saleemrashid:
For reference, see #47 and #48

PR #61, opened 2017-01-22
Add LBRY Credits (LBC)

tzarebczan:
As used here in our small android wallet: https://github.com/fillerix99/lbry-android


prusnak:
Is there a reason why skip to index 150 even when there are unallocated lower numbers?

tzarebczan:
No reason, the original dev just chose 150 knowing it would most likely be available.  I'll submit a PR for his code for the change as well. Thanks for the quick feedback!

prusnak:
Added, thanks

PR #64, opened 2017-02-13
Strip trailing junk after domain

dpn:
Maintain compatibility with the TREZOR password manager by stripping
anything following the domain as well.

### DIFF
--- a/slip-0016/pwdreader.py
+++ b/slip-0016/pwdreader.py
@@ -84,6 +84,12 @@ def getDecryptedNonce(entry):
         item = entry['item'].replace('http://', '').replace('https://', '')
     else:
         item = entry['title'].replace('http://', '').replace('https://', '')
+
+    # The Trezor password manager chrome extension also strips anything
+    # following the first slash in the URL- do the same here to maintain
+    # compatability
+    item = item.split('/')[0]
+
     ENC_KEY = 'Unlock %s for user %s?' % (item, entry['username'])
     ENC_VALUE = entry['nonce']
     decrypted_nonce =  hexlify(client.decrypt_keyvalue(


prusnak:
// cc @chren-de-large

conejoninja:
TREZOR password manager is stripping everything after the / but not after ?
Examples:
https://domain.tld/?id=123 --> Unlock domain.tld for user?
https://domain.tld?id=123 --> Unlock domain.tld?id=123 for user?
In both cases, the protocol is removed

I think we should just take the domain/host part of the url and forget the rest to keep it clean.

Also, in pwdreader.py, only http/https is removed https://github.com/satoshilabs/slips/blob/master/slip-0016/pwdreader.py#L83-L86 while the password manager will remove other protocols (like ftp)

prusnak:
`https://domain.tld?id=123` is not a valid URI but I agree about the rest of the comment

chren-de-large:
hi @dpn , thanks for your contribution & @conejoninja for good explanation. 

Unfortunately, TPM is using external [js lib](https://github.com/chriso/validator.js ) for [url validation](https://github.com/chriso/validator.js/blob/fdea1d79e6ab0fb256f29dd651641dcf7c46012e/lib/isURL.js) and it is little bit more complex than just stripping protocols. Therefore, it would be ideal to rewrite it into the pwdreader or keep it as it is - as proof of concept. Taking the first part after slash can break some other cases ...

prusnak:
Fixed differently in https://github.com/satoshilabs/slips/commit/d9efc929d282d75c2f508abdaac49efa070ac55e

It uses built-in module `urlparse` to try to parse the item as URL. If it wents OK (`pr.scheme` is filled in), let's use the `pr.netloc` field (corresponds to `host:port`).

PR #69, opened 2017-03-23
Add coin type for Decred

coblee:


prusnak:
Please amend the commit to contain uppercase hex digits like the rest of the document.

coblee:
The rest of the document actually has lower case hex. Only the few below my
change has uppercase hex. So which should I change?

prusnak:
Ah, have not spotted this earlier. Will fix in a separate commit. Yours is merged. Thanks

coblee:
Thanks!

PR #71, opened 2017-03-31
Starting segwit

runn1ng:


runn1ng:
I took the segwit addresses and xpub from trezor itself; and the private keys from https://iancoleman.github.io/bip39/ .

Please check the correctness by some third party with some other tools before merging :)

....also the tests are failing because of some markdown details

PR #78, opened 2017-05-18
Add Vertcoin bech32 addresses

metalicjames:
Right now Vertcoin is using `vtc` and `tvtc` for main net and test net respectively.


prusnak:
Thanks!

PR #86, opened 2017-06-22
proposing 111 for ARK

fix:
head of dev of ark here. We are working on hardware wallet integration and this needs to be put in marble :)


prusnak:
Carving it to stone then :-)

fix:
nice, thanks :)

PR #113, opened 2017-09-26
Adding NEO coin to slip 0044

coranos:
NEO is integrating with the Ledger hardware wallet, so we will use coin # 888


prusnak:
Added in https://github.com/satoshilabs/slips/commit/5bf0ca3382b7204657be670b153225ffb9880770

PR #120, opened 2017-10-16
Add Qtum

icodeface:


prusnak:
There cannot be two coins with the same number allocated.

PR #124, opened 2017-11-08
Add Bitcoin Gold to slip-0173

h4x3rotab:
Bitcoin Gold will take `btg` for mainnet and `tbtg` for testnet bech32 addresses.

### DIFF
--- a/slip-0173.md
+++ b/slip-0173.md
@@ -29,6 +29,8 @@ hrp    | coin
 `tltc` | Litecoin Testnet
 `vtc`  | [Vertcoin](https://vertcoin.org/)
 `tvtc` | Vertcoin Testnet
+`btg`  | [Bitcoin Gold](https://bitcoingold.org/)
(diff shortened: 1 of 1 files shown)


jhoenicke:
@prusnak / @clarkmoody: Should the list be kept in alphabetic order?  By name or by prefix?

prusnak:
Let's not sort this for the time being.

clarkmoody:
I agree it's fine for now. Too many more coins and it will need sorting.
Also, should testnet prefixes be kept with their mainnet counterparts? Or
should there be two tables?

prusnak:
Reformatted the table in https://github.com/satoshilabs/slips/commit/97e38962cfdb620aa07a4b73304a87ad21f04320

I think it is better this way ...

clarkmoody:
@prusnak The multi-column edit looks great, thanks.

PR #137, opened 2017-11-23
SLIP-48: Adding ref link to Graphene Wiki

jonathancross:
When reading this SLIP, I had a difficult time locating basic info about Graphene.
Perhaps this link will help others (I also updated the Graphene wiki today, so it's a little more informative).


prusnak:
:+1:

PR #156, opened 2017-12-28
Added entry for Phore

tohsnoom:


prusnak:
Use lowercase for hex values

PR #158, opened 2017-12-29
Suggest changes to slip-0039

dsprenkels:
**DISCLAIMER: These are just some suggestions. I did NOT perform a thorough security review of the methods specified in this document and cannot guarantee that there are no flaws left in this draft.**

Changes that were made:

- Specify explicitly which finite field is used, not only its order but also the irreducible polynomial by which it is reduced.
- Specify the mapping of the key bytes and the finite field in which the secret sharing is computed.
- Why use 2048 PBKDF2 rounds? Why not a round number like 2000/2500/etc.?
- The current version suggest using one round of AES, which would be terribly insecure! Use  at least a full computation.

### DIFF
--- a/slip-0039.md
+++ b/slip-0039.md
@@ -33,12 +33,12 @@ In case sufficient M values are provided the points exactly define the polynomia
 
 ## From entropy to mnemonic secrets
 
-The value to be encoded as the master secret must be a multiple of 8 bits. This is typically a wallet entropy, but we do not restrict the usage in any way. The master secret is divided into N Shamir parts and M specifies how many of those parts do we need to reconstruct the master secret. We use GF(256) as the underlying field. We consider the master secret in a form which includes its own checksum:
+The value to be encoded as the master secret must be a multiple of 8 bits. This is typically a wallet entropy, but may be another secret value which was uniformly chosen from its (key) space. The master secret is divided into N Shamir parts and M specifies how many of those parts do we need to reconstruct the master secret. We use GF(256) reduced by x^8 + x^4 + x^3 + x + 1 (the Rijndael polynomial) as the underlying field. We consider the master secret in a form which includes its own checksum:
 
 | master secret | 16-bit master secret checksum |
 |---------------|-------------------------------|
 
-From this value the N parts are generated and each participating party receives the following data:
+From this value, every byte is mapped to the specified field in a little-endian fashion (i.e. the first bit maps to a_7, the last bit maps to a_0). For each field element (S), the N share-elements are generated and mapped back to bytes. Each participating party receives the following data:
 
 | 5-bit index | 5-bit M threshold | variable-bit SSSS part | 16-bit checksum |
 |-------------|-------------------|------------------------|-----------------|
@@ -69,17 +69,18 @@ This structure is then converted into a mnemonic passphrase by splitting it up b
 
 ## Checksum
 
-For checksum we use 16-bits of SHA-256 hash function.
+For the checksums we use the leftmost 16 bits of a SHA-256 hash digest of the relevant payload.
 
 ## Passphrase
 
 When enough M secrets are provided the master secret is reconstructed. To allow an additional protection of the final seed using a passphrase we suggest a following scenario. The passphrase is stretched using PBKDF2, which servers as a key to an AES encryption. The result of this encryption is the final seed to be used. If no passphrase is provided the passphrase is set to an empty string.
+(TODO: Use domain separation in PBKDF2)
 
 ![passphrase](slip-0039/passphrase.png)
 
 This mechanism is intentionally symmetric to allow reconstructing the master secret from the seed in case the passphrase is known.
 
-We propose to use 2048 rounds of PBKDF2 and 1 round of AES (TODO: which mode?).
+We propose to use 2000 rounds of PBKDF2 and one full computation of AES (TODO: which mode?).
 
 ## Versioning
 


prusnak:
> Why use 2048 PBKDF2 rounds? Why not a round number like 2000/2500/etc.?

We used 2048 for older BIP39 standard. This number was inspired by WPA which uses 4096 rounds of PBKDF2 (hence the number of two of iterations).

> The current version suggest using one round of AES, which would be terribly insecure! Use at least a full computation.

Yes, that's what I mean but one round - the full iteration.

dsprenkels:
> Yes, that's what I meant by one round - the full iteration.
I guessed so. ;)
But as it was written, it could be interpreted either way.

slip-0039.md
```diff
@@ -69,17 +69,18 @@ This structure is then converted into a mnemonic passphrase by splitting it up b
...
 
 When enough M secrets are provided the master secret is reconstructed. To allow an additional protection of the final seed using a passphrase we suggest a following scenario. The passphrase is stretched using PBKDF2, which servers as a key to an AES encryption. The result of this encryption is the final seed to be used. If no passphrase is provided the passphrase is set to an empty string.
+(TODO: Use domain separation in PBKDF2)
 
 ![passphrase](slip-0039/passphrase.png)
 
 This mechanism is intentionally symmetric to allow reconstructing the master secret from the seed in case the passphrase is known.
 
-We propose to use 2048 rounds of PBKDF2 and 1 round of AES (TODO: which mode?).
+We propose to use 2000 rounds of PBKDF2 and one full computation of AES (TODO: which mode?).
```
saleemrashid: Which PRF for the PBKDF2? I assume it is HMAC-SHA512.

slip-0039.md
```diff
@@ -69,17 +69,18 @@ This structure is then converted into a mnemonic passphrase by splitting it up b
 
 ## Checksum
 
-For checksum we use 16-bits of SHA-256 hash function.
+For the checksums we use the leftmost 16 bits of a SHA-256 hash digest of the relevant payload.
 
 ## Passphrase
 
 When enough M secrets are provided the master secret is reconstructed. To allow an additional protection of the final seed using a passphrase we suggest a following scenario. The passphrase is stretched using PBKDF2, which servers as a key to an AES encryption. The result of this encryption is the final seed to be used. If no passphrase is provided the passphrase is set to an empty string.
+(TODO: Use domain separation in PBKDF2)
```
prusnak: What exactly do you mean by this?

saleemrashid: @prusnak Presumably some sort of salt

slip-0039.md
prusnak: I think it's good idea to use HMAC-SHA256 as SHA512 is much faster on 64-bit processors giving a desktop computer fair advantage over 32-bit embedded ARM. I'll have to figure out the number of rounds though.

saleemrashid: I assumed SHA512 because that's what BIP39 uses. Exactly why we need to specify it :wink:

prusnak: Actually, I wanted to use SHA512, but your comment got me thinking more about that :)

prusnak:
Thank you, Daan! Merged

slip-0039.md
dsprenkels: Yeah, so I am not really sure about this. My gut tells me it would be a good idea to include a salt here, to make it secure against rainbow-table attacks. I am not really sure though.

saleemrashid: @dsprenkels ACK regarding a salt. We're only using this as a symmetric encryption key, so it cannot hurt to have a salt since the user can still retrieve the master secret and switch to another standard, if they wish.

slip-0039.md
saleemrashid: Then it might be worth considering using another, more ASIC-resistant key stretching algorithm. We could even use multiple algorithms (for example, PBKDF2 and Scrypt or Argon2, which have high memory requirements)

prusnak: At the same time you want to make this computable on TREZOR, so Scrypt and most probably Argon2 too are out of the question.

slip-0039.md
prusnak: Included fixed salt in the spec for now.

slip-0039.md
saleemrashid: Realized that just after I commented. :wink:

But I will take a look at other algorithms such as Bcrypt. Assuming there are no side-channel attacks, adding another algorithm will make it at least as secure as a single algorithm.

slip-0039.md
tsusanka: I am also not 100% sure on this, but I believe we need salt in case the master secret is some kind of predictable value (like passwords or dictionary words), but it is not required in our case where the master secret is the wallet entropy, which is required to be random on its own.

One way or another, the PBKDF2 requires salt anyway and it definitely doesn't hurt to use it. In the traditional case of storing passwords in a database, you should use per-password unique salts to prevent mass computation of all the passwords. Since this is not our case I believe using a hard-coded string as a salt is fine. Correct me if I'm wrong.

saleemrashid: > I believe we need salt in case the master secret is some kind of predictable value

@tsusanka If that's the case, bear in mind that this standard could be used for most kinds of data (including those which may not be completely random).

tsusanka: @saleemrashid you're right. Giving it another thought the salt being random would be the best scenario, but a user would have to remember some additional random data and I'm not sure we want to go this way. I have to think more about this :)

tsusanka: @saleemrashid I think I'm wrong, it doesn't make sense. Is it okay if I delete our comments?

saleemrashid: @tsusanka It's up to you :smile:

PR #160, opened 2018-01-04
Add NewYorkCoin, GoByte, BitcoinZ, Poa, Omni

erasmospunk:


saleemrashid:
Shouldn't Omni use the Bitcoin private keys? You need BTC in order to use OmniLayer (for transaction fees).

prusnak:
@erasmospunk can you answer? leaving this PR unmerged creates new conflicts as new numbers are being allocated every day

erasmospunk:
@saleemrashid @prusnak the Omni Layer is account based like Ethereum and it reuses the same key. This is the same as in Counterparty.

saleemrashid:
@erasmospunk I understand that, but how will users pay fees? They need BTC (or LTC).

saleemrashid:
For that matter, how will you distinguish between BTC and LTC OmniLayer?

PR #162, opened 2018-01-06
Number correction

dabura667:
11111 is 31
100000 is 32 (2 ^ 5)


prusnak:
Read the whole table again, the values are shifted by one.

PR #181, opened 2018-01-30
Added Koto

kotodev:


prusnak:
Use lowercase letters for hex string like rest of the table.

PR #188, opened 2018-02-02
Added HOdlcoin

FreeTrade:


prusnak:
Use lowercase letters for hex values

PR #193, opened 2018-02-06
Add SAFE

github10000:
Add  SAFE to BIP-0044


prusnak:
6688 is not 0x80006688, fix the hex value

PR #196, opened 2018-02-08
Add QTUM

superbigfu:
Signed-off-by: bepal

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -243,6 +243,7 @@ index | hexa       | symbol | coin
 1524  | 0x800005f4 |        | [Taler](http://taler.site)
 1815  | 0x80000717 | ADA    | [Cardano](https://www.cardanohub.org/en/home/)
 1989  | 0x800007c5 | HODL   | [HOdlcoin](https://hodlcoin.com/)
+2301  | 0x800008fd | QTUM   | [QTUM](https://qtum.org/en/)
 4242  | 0x80001092 | AXE    | [Axe](https://github.com/AXErunners/axe)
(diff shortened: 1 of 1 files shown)


icodeface:
傻逼？

icodeface:
hi @prusnak, Qtum doesn't use 2301 as coin type

prusnak:
That's not good. The document says 2301 so all wallets shoulf use this
value. If there are some wallets which use wrong number, notify them to
change it to the correct one.

prusnak:
I can remove the line, but 88 is already used by another coin.

icodeface:
@prusnak thanks

fanatid:
http://exodus.io/ already uses `2301`

dooglio:
Please hold off on reverting anything. I've contacted the Qtum dev team to find out what their stance is.

fanatid:
@prusnak even if decision will remove `2301`, I think that this number should stay in table and other coins should not take it

qtum-neil:
Please remove 2301 as it was not made by a Qtum team member, and it does not reflect the real number used in our wallets, Qtum uses 88 as constant, an earlier request was made here: https://github.com/satoshilabs/slips/pull/120 but it was closed due to there being a duplicate.

I think duplicates should be allowed when they reflect reality.

fanatid:
@qtum-neil thanks for confirmation :(

@prusnak can we mark `2301` as not available and put link to this issue?

dooglio:
Can we at least add a remark into the `88` entry to show that it's also being used by Qtum?

Issue #195, opened 2018-02-07
SLIP Request: Registered version bytes for BIP-32 HD seed serialization


clarkmoody:
As with [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) (registered coin types) and [SLIP-173](https://github.com/satoshilabs/slips/blob/master/slip-0173.md) (registered human-readable bech32 prefixes), we need some place to catalog [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format) version bytes. For Bitcoin mainnet, this is `0x0488B21E` which produces the `xpub` prefix on HD public seeds when Base58 encoded.

Even @prusnak's [BIP-84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki) fails to document the version bytes needed to reach `zpub` in the HD serialization. I did find [a post on the mailing list](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-September/014907.html) proposing the `xpub`, `ypub`, `zpub` version bytes, but I don't think this is recorded anywhere else except the source code of wallets. (Please correct me if I'm wrong on this count)

The issue I've found is that multiple values of the version bytes will produce the same ASCII prefix when Base58 encoded. It would be nice to maintain a registry of those version bytes here.

As a side note: I'm hopeful that the adoption of [SLIP-32](https://github.com/satoshilabs/slips/blob/master/slip-0032.md) encoding of HD seeds will make this request redundant, but I don't see wallet adoption happening soon.

prusnak:
It is documented in the Electrum documentation, which is also linked from the post on the mailing list you mention.

http://docs.electrum.org/en/latest/seedphrase.html

As this documentation is Bitcoin-centric, I agree we might start a SLIP for non-Bitcoin coins, but currently, I don't have a capacity to do that. 

Clark: would you like to start a document like this and submit it to this repository?

clarkmoody:
Of course I knew you'd say that :-)

Since we already have SLIP-32, what number should this SLIP be?

prusnak:
No strong opinion. 33, 132 or any other fitting number you'll come up with.

clarkmoody:
Closing due to merge of #197

PR #197, opened 2018-02-08
Add SLIP-0132: Registered HD version bytes for BIP-0032

clarkmoody:
Addresses #195 by adding all known Bitcoin HD version bytes, Litecoin, and Vertcoin.


prusnak:
Great work!

prusnak:
Added two rows in f83f034f50461d8bbadee96bf4da359336d95eeb

PR #199, opened 2018-02-11
Update slip-0044.md

fachain:


prusnak:
Please respect the order and insert the line where it belongs, not at the end.

PR #222, opened 2018-03-10
SLIP-0044: Add NEETCOIN(NEET)

shirako-jp:


tofutim:
How does one register to add a new coin?

prusnak:
@tofutim send a pull request

PR #234, opened 2018-03-27
Requesting 1729' for Tezos

murbard:
Tezos currently uses ed25519 and is implementing bip32-ed25519 (which may use a different nomenclature). However, it does plan on using secp256k1 signatures as well so, all-in-all, it makes sense to reserve a bip44 constant.


prusnak:
Something is not right. 1729 != 4ff

murbard:
I accidentally 1279. Fixed in the second commit.

PR #235, opened 2018-03-28
Update slip-0044.md

WaykiChain:
add waykichain


prusnak:
Use lowercase for hex constants like the rest of the documents

PR #261, opened 2018-05-07
Add EAST support

eastcoinuser:
Eastcoin is an ethereum fork coin for travelers ( See more info on http://easthub.io)

We would like to reserve 625 as a chainid / networkid / hwpath m/44'/625'/0'/0/0


prusnak:
Why do you mention 625 and then use 1987 in the address path?

eastcoinuser:
@prusnak 0x8000022b -> 555, 0x80000271 -> 625

Or should I use another path?

eastcoinuser:
@prusnak Didn't checked the hwpath sorry 👍

PR #263, opened 2018-05-10
Add Ethereum Social support

akx20000a:
Ethereum Social is a Community enterprised blockchain platform based on ethereum. ( See more info on https://ethereumsocial.kr)

We would like to reserve 1128 as a chainid / networkid / hwpath m/44'/1128'/0'/0/0.


prusnak:
Testnets have the same value (which is 1), so please remove the testnet line.

akx20000a:
@prusnak Removed, Thanks!

PR #288, opened 2018-06-15
Add stacks coin-type

kantai:
Stacks are the upcoming token on the blockstack network (testnet at https://testnet.blockstack.org/)


prusnak:
Use lowercase letters for hex value like the rest of the table does.

kantai:
Updated. Sorry about that

PR #293, opened 2018-06-20
Update slip-0044.md - Include Teslacoin

TeslacoinFoundation:


prusnak:
@TeslacoinFoundation The values you provided (decimal and hex value) do not match. Which one is the correct?

PR #294, opened 2018-06-21
Improve SLIP-0010 readability

webmaster128:
This is an idea to make the SLIP a bit more readable for people, that did not yet implement it. For me it was not clear at all what an "fpr" should be. Additionally it helps to differentiate between the chain (e.g. `m/0_H/1/2_H`) and the chain code (e.g. `873dff81c02f525623fd1fe5167eac3a55a049de3d314bb42ee227ffed37d508`).


prusnak:
Thanks!

PR #295, opened 2018-06-21
SLIP-0010 "chain" ->  "path"

webmaster128:
* change label "Chain" -> "Path"
* change path root `m` -> `/`

as described in the commit messages


prusnak:
I don't agree with second two commits of the change - see https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#test-vectors

First one is fine, so please rework the PR to include only the first one.

webmaster128:
Alright, updated. Thanks for feedback. I'll think about it for some more time

prusnak:
Thanks!

PR #296, opened 2018-06-21
SLIP0010: fix different expressions for reulting child key

webmaster128:
Before, we checked

* `I_L` for >= n
* `I_L + k_par` for 0

and returned `I_L + k_par (mod n)`.

Both those 3 expressions should be the same in order to avoid returning a key >= n or equal to 0.


webmaster128:
Ah, sorry. The in the patch the definition of k_i is missing.

webmaster128:
check

webmaster128:
For some reason, BIP0032 also states that `I_L ≥ n` makes `I_L + k_par (mod n)` invalid. I don't see why this should be the case but I'll update this to match BIP0032.

webmaster128:
Oh, the failure in case of `I_L ≥ n` comes from a time when `I_L` was multiplied, not added: https://github.com/bitcoin/bips/commit/d0ae2739773ef6ac476f514a6f99920da1882ca6

webmaster128:
I reverted the deduplication of k<sub>i</sub> in order to make the diff as small as possible for an easy review

prusnak:
Thx!

PR #314, opened 2018-07-09
New coins and table format

clarkmoody:
Adds DigiByte, Ravencoin, Namecoin

Reformats table with one column for coin and three columns for mainnet, testnet, and regtest prefixes.


prusnak:
Thanks!

PR #316, opened 2018-07-12
[SLIP-0173] Update testnet + regtest bech32_hrp for Bitcoin Private

ch4ot1c:
For testnet, `test` should be used instead of `t`; Zcash/Zclassic base58check addresses begin with `t`, which could confuse users.

For regtest, `rtest` has been added.

https://github.com/BTCPrivate/BTCP-Rebase/pull/30


prusnak:
Actually the Zcash addresses start with `t1`, not `t`. So it's not confusing. I change the SLIP-0173 values to match the current master of BTCP (before your pull request to their repo) and I am closing this one.

ch4ot1c:
It's confusing to end-users who don't understand alphanumeric mode vs base58check, and only remember/consider a single prefix character (wrongly). Glad it's changed.

PR #317, opened 2018-07-12
new Slip: Encoding network identifier in Bip44 paths

dan-da:
This is my first attempt at writing a Slip.

I temporarily chose Slip-0144 as it is related to both Bip44 and Slip44, but that can be changed to anything.

I hope the idea will be accepted, or at least to spark discussion of the issue and merits of this proposal.

### DIFF
--- a/slip-0144.md
+++ b/slip-0144.md
@@ -122,7 +122,6 @@ integer value.
 
     altnet_idx - (network_index * 100000000) = coin_idx
 
-
 ## Calculation Examples
 
 ### Calculating Altnet Index from Mainnet Index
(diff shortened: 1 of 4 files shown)


prusnak:
TBH I don't like this. coin_type of 1 is not meant to be used just for Bitcoin Testnet, but all coin Testnets including Bitcoin regtest. The main reason for this separation is that testing keys won't mix with production keys.

If a coin has more production chains, it should register a bip44 value for each chain. Most of the chains are for testing, however, and they can happily use coin_type = 1

dan-da:
@prusnak thanks for your comments.  I'm disappointed but I guess not really surprised you don't like it as you are an author of Bip44 which would have been the logical time to include coin_type/network in the path, and Slip44 which conflated testnets for all coins together into one value, which loses information.  I guess I was hoping it was just an oversight though.

Surely though you can recognize that it is cleaner for software that deals with keys to be able to treat bip32 paths and keys the same (identical codepaths) regardless of whether they are Mainnet keys or testnet keys.  Separate codepaths means greater chance of bugs.

In the case of Slip32, which I support and I hope finds adoption, the path is encoded in the key.  This means that software implementing Slip32 can look at a given key and determine the coin.  Ok, great.  Now it can do something with it, even if that is just presenting "This key was issued for Y coin", or maybe it is something more useful like looking up Tx's on a blockchain,  But if it has coin_type=1, then it cannot know which coin, and must follow a different code path, which is typically going to be something lame like "I don't know which blockchain this is for, sorry I give up".   So here we see software having to treat testnet keys as second class citizens, and this particular codepath cannot be tested for a testnet key -- which sort of negates their usefulness for testing.

At a fundamental level to me this is a very simple matter:  why lose information about the coin_type when (a) that is unnecessary and (b) the entire point of the field is to convey the coin_type.  I can't fathom it.

Stated another way: In my view, assigning value=1 to all testnets for all coins was a corruption of the meaning of coin_type, resulting in a lossy encoding.  Maybe not against the wishes of everyone because most people won't care one way or the other, but certainly against my wishes. Speaking as an author of a tool for deriving and generating hd wallet keys, including testnet keys, for 300+ coins... I want a rock solid method of identification..

So my proposal is simply a method to preserve both the coin and the network in the encoding.  It won't affect software that doesn't care about testnets, but it will allow software that DOES CARE to encode and decode the information it needs. 

If you or anyone else have another proposal that will achieve the same goal, I'd be happy with that.  But I do not find the testnets "can happily use coin_type = 1" argument compelling, nor the idea that testnet keys and production keys "won't mix".  of course they will and do.  They mix all the time right here on my laptop.  It is better that we can properly identify and work with them.

So, I hope you will reconsider...

prusnak:
> Speaking as an author of a tool for deriving and generating hd wallet keys, including testnet keys, for 300+ coins... I want a rock solid method of identification..

I guess this is the main confusion. The coin_type field is to separate the keys to provide security accross the various chains, not to provide rock solid method of identification.

dan-da:
"across the various chains", exactly.  coin+network uniquely identifies a blockchain.  coin alone does not unless restricted to Mainnet only...  which Bip44 and Slip44 do not do.  Hence.... these promote confusion.

ManfredKarrer:
I support the idea of @dan-da.

Hoangthienbao:
Toi hieu

PR #321, opened 2018-07-20
zPIV derivation path added

furszy:


prusnak:
pivx is already registered as 119

furszy:
This is not about regular PIV, this is about zPIV. Guess that you are aware about the zerocoin implementation that we have.
The thing is: Most of the wallet don't support zPIV, basically because of lack of knowledge etc etc. The public part associated to the coin is a Pedersen commitment and the key generation is not standardized outside of the core wallet.
What I want here is standardize the zPIV in a different path or will have to fix an specific account number for this use case. The derivation path cannot be the same as the regular PIV.

prusnak:
Ok, please pick another value or you might hit some problems with some wallets.

furszy:
Done.

PR #332, opened 2018-07-31
fix and add new coin_index

superbigfu:
fix coin_index 2303 symbol

add 2304  | 0x80000900 | SSC    | [SelfSell](https://www.selfsell.com)


prusnak:
@fanatid already fixed in https://github.com/satoshilabs/slips/commit/ae1816bc96acbb6dd98395dded3c271cb735542b yesterday

PR #340, opened 2018-08-08
Request to add 0x800000fd NPW https://npw.live

guantau:
Request to add 0x800000fd NPW https://npw.live


guantau:
It seems that line 287 has one extra space at the end.

PR #344, opened 2018-08-17
add Upub and Uprv

jlopp:
I noticed that the SLIP does not have an entry for P2WSH in P2SH for TBTC. I'm adding the version bytes that we are using at Casa based upon the entry in SomberNight's draft BIP at https://github.com/SomberNight/bips/blob/bip-xpub-versionbytes-extension/bip-xpub-versionbytes-extension.mediawiki


prusnak:
Thanks!

PR #364, opened 2018-09-15
Add Monacoin bech32 addresses

wakiyamap:
Monacoin currently uses the human readable part below.

mainnet `mona` https://github.com/monacoinproject/monacoin/blob/master-0.16/src/chainparams.cpp#L147
testnet `tmona` https://github.com/monacoinproject/monacoin/blob/master-0.16/src/chainparams.cpp#L256
regtest `rmona` https://github.com/monacoinproject/monacoin/blob/master-0.16/src/chainparams.cpp#L363


wakiyamap:
thanks!!

PR #388, opened 2018-10-04
SLIP-0039 Use the PMS hash in the leading coefficient of the Shamir polynomial.

andrewkozlik:
Use the PMS hash in the leading coefficient of the Shamir polynomial.
Add a detailed description of how the share values are generated.
Remove the PBKDF2-only proposal from the list of master secret derivation functions.


prusnak:
Merged to master, thanks!

PR #418, opened 2018-10-28
Adding NRG - Energi code

andvgal:
The code is based on communication port used - 9797.


prusnak:
Good find!

@andvgal NRG was already assigned code number 204. Use that

andvgal:
@fanatid @prusnak - the code 204 has never been used, but 9797 is used in production.
I am not sure who is the original author and why did he make the request on behalf of the project.

Please get the change back. Otherwise, the SLIP just does not contain the actual information.

prusnak:
@andvgal You should assign number in SLIP44 before putting the code into production. Send another PR to change it to 9797.

andvgal:
@prusnak The registration has been done on Oct 28, the release has been done on [Nov 6](https://github.com/energicryptocurrency/energi/releases), the SLIP change got reverted in Nov 10  without any consultation with the project. 

OK, you'll get another PR, but please stop blaming with false accusations.

prusnak:
@andvgal The original assignement happened on May 10th - see https://github.com/satoshilabs/slips/pull/264 so the accusation is true.

PR #434, opened 2018-11-08
Update slip-0044.md with NULS main chain ID

moshemalawach:
NULS uses the secp256k1 curve and has a chain_id of 8964 for main net chain.

For the record, I am one of the core community developers of NULS.


moshemalawach:
I am open to any questions, thanks in advance :)

PR #441, opened 2018-11-12
add Genom

Gy9vin:


prusnak:
@Gy9vin The values you provided (decimal and hex value) do not match. Which one is the correct?

Gy9vin:
> @Gy9vin The values you provided (decimal and hex value) do not match. Which one is the correct?

the project closed in December 2018

PR #442, opened 2018-11-12
add Blockstream's Liquid Bitcoin sidechain token

farmerwampum:


prusnak:
@greenaddress Does liquid need a separate SLIP44 constant?

prusnak:
@farmerwampum 231 is taken, choose another value

PR #443, opened 2018-11-13
slip-0044: Adding QRL

jleni:
Adding Quantum Resistant Ledget to SLIP-0044


prusnak:
Don't change the row 237

PR #448, opened 2018-11-16
Register new currency BitcoinVIP

snight1983:
Register new currency BitcoinVIP
520   | 0x80000208 | BTCV   | [BitcoinVIP](https://www.bitvip.org/)


snight1983:
Register new currency BitcoinVIP
520   | 0x80000208 | BTCV   | [BitcoinVIP](https://www.bitvip.org/)

PR #462, opened 2018-11-27
Add IPC in slip-0044.md

IPCChain:
Add in slip-0044.md


prusnak:
So which one is the correct value you want to assign? Please send a new pull request with one agreed value.

IPCChain:
Hello prusnak, i want to assign 0x800000ec  as IPC, from my fork file, nobody use this value, could I use this value?

PR #468, opened 2018-12-03
Energi / NRG changes

andvgal:
- Restored correct Energi Mainnet 9797 registration
- Added Energi Testnet 19797 registration
- Removed never used 204 registration


prusnak:
You should use 1 for testnet.

fanatid:
Please left 204 as not available for registration, because this constant was used.

prusnak:
@fanatid used where? Can you show me the wallet which uses 204?

fanatid:
@prusnak uses in SLIP-044 is not enough? i.e. is assign some number and then deassign -- right?

prusnak:
If there is no wallet using 204 for NRG I would like to keep it available.

andvgal:
> You should use 1 for testnet (all coins). Remove the 19797 value.

A separate value is used on purpose as Energi supports both legacy (sharing Dash code) and new HD wallets. The different value is used for testing purposes in regtest/testnet.

There is at least one more coin having a separate testnet ID (37310). So, the requirement to remove one for Energi would be discriminatory.

prusnak:
Meh

PR #482, opened 2018-12-16
add OKChain

zccst:


prusnak:
use abbreviation in abbreviation column, not the full name

zccst:
> use abbreviation in abbreviation column, not the full name

sorry, that's okp

PR #493, opened 2018-12-29
slip-0173 add BNB

rickyyangz:
This is for the binance chain


prusnak:
1) binance.org does not work
2) can you point me to the source of this coin?

rickyyangz:
The Binance Chain is still under development but close to launch. After launch, binance.org would be open and the source code would be gradually open sourced.

prusnak:
OK, reopen if the coin is launched and has a public source code.

rickyyangz:
is it a must to wait for the coin launched and open sourced? actually, we have added BNB in slip-0044.
@prusnak

prusnak:
How am I supposed to check the validity of the values if the source is not available?

PR #494, opened 2018-12-29
[slip-0044] Add Snowblossom Coin Type(s)

cluelessperson:
Thanks


prusnak:
All testnets should use the same constant: 1.

Also I am not sure if it makes sense to use BIP44 for channels. My take is that you should use a completely different path for these.

prusnak:
Closing. Reopen if you are alive again and willing to fix this according to my instructions.

PR #510, opened 2019-01-16
[slip-0044] add url for binance chain

ackratos:
We finally decide we will use www.binance.org for our binance chain DEX, a follow up of https://github.com/satoshilabs/slips/pull/424


ackratos:
@prusnak thanks for merging!

PR #519, opened 2019-01-28
add swtc coin

caivega:


caivega:
add swtc coin at 315   | 0x8000013b | SWTC   | [SWTC](http://swtc.top/)

caivega:
The SWTC Public Chain has operated successfully for many years since the launching on Oct 2014. It applies the self-developed RBFT consensus mechanism with 5-layered underlying technology to support commercial applications, 
The SWTC Public Chain has reached more than 10 million blocks height (until August 2018). In addition, the SWTC Public Chain supports cross-chain function - can be connected with micro chain and other commercial consortium or private chain.
The vision of SWTC Public Chain is to provide a safe, valid, and credible commercial blockchain environment. At the same time, The SWTC Public Chain is also positioned to be a decentralized trading platform that accommodates a variety of digital assets.

Issue #315, opened 2018-07-09
Slip132: doesn't specify what scripts the script-hash types are for`


afk11:
https://github.com/satoshilabs/slips/blob/master/slip-0132.md

I noticed in a discussion recently that this is meant to follow electrum in the prefixes used. I had made the mistake of assuming the P2SH, P2WSH, P2SH|P2WSH types referred to P2PKH inside the respective script type, but no, that it was for multisig.

Could the document be updated to reflect that they refer to multisig? I started implementing P2PKH inside these types as 'slip132' assuming they all referred to key hash types.

afk11:
*P2SH*, *P2WSH* and *P2WSH in P2SH* aren't definite script types, they just say that some script type is wrapped in some form. 

It not specifying the type of the redeem/witness script is the confusion here, not what the letters stand for.

prusnak:
I get your point now. If you have a particular idea how to improve the document, please, send the pull request.

afk11:
@prusnak I submitted a pull request regarding this issue. It only modifies the Zp & Yp prefixed keys, as I have yet to encounter xpub/xprv prefixes used in a P2SH context. Is there anywhere that actually makes use of that, and what script type have they nested inside P2SH?

PR #531, opened 2019-02-13
SLIP 132: Mention multisignature script type for Yp and Zp prefixed keys

afk11:
Late follow up, but hope this is an acceptable fix for https://github.com/satoshilabs/slips/issues/315

Really don't like handling these as multisig without being able to refer to the spec.


clarkmoody:
The xpub encoding story continues to be a disaster. I'm not sure we've made any headway over the past 18 months, expect the draft SLIP-32?

Tracing through the Electrum source, this looks like what they're doing, so it's an ACK from me. It would be much more helpful to have a BIP describing the address encoding scheme?

afk11:
@clarkmoody thanks for review!

As far as I can tell, the P2SH usage in 'p2pkh or p2sh' is also multisignature. Would it be ok to submit this also?

Also, I forgot to apply the changes to other networks - I'll follow up with that now!

clarkmoody:
I would re-order the rows of Testnet to match the mainnet rows (put multisig rows together).

afk11:
@clarkmoody good catch, I've pushed this now

clarkmoody:
👍 LGTM

CC: @prusnak

prusnak:
Thx!

PR #540, opened 2019-02-25
Adding ZBUX(Z-bux)

interbiznw:


prusnak:
Use the first free entry

interbiznw:
@prusnak Sorry about that, Wasn't sure since I seen them all over the place. 

Moved it to first open slot now, thanks!

PR #546, opened 2019-02-27
Update slip-0044.md

bc-chain:
add BCChain index


prusnak:
Use the first free entry.

bc-chain:
Change index from 5921 to 268

PR #551, opened 2019-03-03
add coin type of shitcoin

furusake-s:
A new PoW coin launched 2019, Shitcoin needs coin type of slip44.


prusnak:
Please use the first free number

furusake-s:
changed

PR #554, opened 2019-03-04
Add a comment for Qtum

icodeface:
the previous discussion is here https://github.com/satoshilabs/slips/pull/196,
since we cannot change the number 88 in our code, and 2301 is also used by some wallets,  leaving a comment here is better.


prusnak:
Official number is defined here and thus 2301 is the official number.

PR #567, opened 2019-03-18
Add BRAVO 0x85425256

bravoadam:
Added BravoCoin BRAVO 0x85425256 bravocoin.com


prusnak:
Use the first available number

PR #569, opened 2019-03-18
add BitZeny in slip-0173

fujicoin:
Please add BitZeny.


fujicoin:
@prusnak Thank you. 
However, because there is objection in the community, there is a possibility to change in the near future.

cryptozeny:
nice job. thanks 😀

PR #579, opened 2019-03-29
Added LUX to the list

giaki3003:


prusnak:
Use the first free number

giaki3003:
> Use the first free number

Hello, is it absolutely necessary? We're about to roll out a few updates out which revolve around this number change

I'm sorry if this caused any issue

PR #582, opened 2019-04-02
XDN digitalnote 2.0

skironDotNet:


prusnak:
use the first free number

skironDotNet:
the rules has changed? I submitted like 5 coins and was never a problem

prusnak:
No, it was always like this

skironDotNet:
No way, look at those all custom selections
custom sel

Also you make invalid statement "was always like this" against other 100 coins with custom HD key reservation
how is this in order

I don't have a problem with "use the first free number" but want to know why now and not before?

prusnak:
> I don't have a problem with "use the first free number"

Please do what I say then

skironDotNet:
Done first available

PR #584, opened 2019-04-04
Update slip-0044.md with BOLD coin

theboldproject:
Registered coin type for BOLD coin.


prusnak:
use the first free numbe

theboldproject:
Updated to first free number

PR #586, opened 2019-04-09
register USE coin to bip44

ablozhou:
register USE coin to 293, https://usechain.net


prusnak:
Use the first free number

PR #587, opened 2019-04-10
slip-0044: add ASF

develnk:


prusnak:
use the first free number

develnk:
Updated.

prusnak:
Use the number 296

PR #592, opened 2019-04-11
slip-0044 : added IOCoin

iocrypto:


prusnak:
Use the number 295

iocrypto:
updated to 295

PR #601, opened 2019-04-17
Adds DeVault (DVT) to slip-0044.md

proteanx:
https://github.com/devaultcrypto/devault


prusnak:
use the first free spot

proteanx:
339 is what we're using already. I feel like it's pretty close to the rest of the list (we made sure to not be unreasonable). Seems like in other PRs (with #s further from list) you were only concerned with correct formatting? Hopefully we can squeeze this one in as it's not like we're asking for '7825267' like obsr etc...

prusnak:
please provide a website with working HTTPS

proteanx:
done

PR #608, opened 2019-04-25
Kuva coin types

koirider:
5 coin types for Kuva stablecoins


prusnak:
You don't need separate values for all the coins. Send just the one entry. And use the first free value.

koirider:
Hello Pavol,

could you elaborate on your point that we don't need separate values?
To add some context: our blockchain is not derived from Ethereum and the coins we need to register are not ERC20 tokens. That implies that each coin actually needs a separate coin type in order to be handled properly by a compatible wallet.

In regards to using first free value. I understand that for you as a maintainer it must be pretty annoying that people randomly pollute coin type space, but there does not seem to be a strict rule for this. In our blockchain implementation we needed a separate continuous address space for a bunch of coins that we need to operate. That's why we applied for sequential numbers "somewhere where it's not too crowded" (3003 registration at the end of March came as a bit of surprise, around which we'll have to dance somehow).

prusnak:
Does each coin have it's own blockchain?

koirider:
The answer to your question is not binary. We plan to have multiple coins on the same instance of blockchain. How is this relevant to the topic of reserving coin type values?
According to how I read the standards, there's no explicit assumption that coin type has a 1-1 mapping to the blockchain instance. It's about keys management rather than blockchain itself. Regardless of how we store the coins / transactions (whether on a single blockchain or in multiple blockchains), each coin needs to have its own coin type in order for a compatible wallet to be able to work with it (i.e. setup the keys and do all the derivations).

prusnak:
The purpose of SLIP-0044 is to isolate different environments. I.e. to protect an user in case some shitcoin privkey is compromised so they will not loose their Bitcoins. All Kuva coins shared the same environment, thus they should share the same index. You can register one SLIP-0044 path for all Kuva coins. Pick a number and we'll assign it. But I am not adding multiple entries for all your coins.

PR #611, opened 2019-04-28
Update slip-0044.md BCZ

devcon-gh:


prusnak:
Show me the proof the coin was indeed renamed

PR #613, opened 2019-04-29
slip-0044: add STASH

BeachM:


prusnak:
use the first available value

BeachM:
@prusnak sincere apologies for the out of order index. Is it possible to make an exception? in this case The reason I ask is that we have live wallets deployed already using current index.

prusnak:
Meh, it was irresponsible to start using the value without allocating it first.

BeachM:
Agreed. Lesson learnt for future. Thanks again.

PR #640, opened 2019-05-29
slip-0044: Add Own (CHX)

marian-gheorghe:


prusnak:

anilmujagic:

Hi @prusnak 

Our wallet app is already deployed and wallets are being created. We understand that you would prefer to have the next available number used, but we didn't infer that from the current entries and didn't think it's required.
Changing it now would create too many issues on our side and would render peoples wallets invalid.

prusnak:
Meh

PR #641, opened 2019-05-29
Update slip-0044.md with ASK

seanmcgary:


prusnak:

seanmcgary:
We're already actively using index `2221` with our mainnet blockchain (and existing wallets), so we'd like to keep it if possible.

prusnak:
Meh

PR #648, opened 2019-06-03
Add USDP & HTDF.md

bluesky0125:
Add USDP & HTDF


bluesky0125:
We want to add USDP & HTDF with the same index with Bitcoin Cash(index 145).

prusnak:
This is not possible. Change your index to a different number.

PR #650, opened 2019-06-03
add cpc

cpchainbot:
Pls review this pr,
one more question: is this coin_type different from chainid? What do we need to pay attention to this coin_type before our main-net release? We just want to be supported by some cold wallet.


prusnak:
Please edit the format to match other entries

slip44 is different to chainid

prusnak:
Try again, still one more space is missing after "`|`" and before "`[`"

yuppieghost:
> Try again, still one more space is missing after "`|`" and before "`[`"

ok, sorry about that.

PR #662, opened 2019-06-15
Create slip-0077.md

romanz:
Following https://github.com/trezor/trezor-firmware/issues/244.

### DIFF
--- a/slip-0077.md
+++ b/slip-0077.md
@@ -24,7 +24,9 @@ blinding_private_key := HMAC_SHA256(key=master_blinding_key, msg=script_pubkey)
 blinding_public_key := secp256k1_publickey(private_key=blinding_private_key)
 ```
 
-The receiver is using `blinding_public_key` construct a "blinded address", which is used by the sender to blind the relevant transaction outputs. Each such blinded transaction output also contains the sender's ECDH public key, so the receiver would be able to recover the shared nonce using its `blinding_private_key`.
+Note: `blinding_private_key` (as 256-bit scalar) must be less than the secp256k1 curve group order - otherwise, the derivation above must fail.
+
+The receiver is using `blinding_public_key` to construct a "confidential address", which is used by the sender to blind the relevant transaction outputs. Each such blinded transaction output also contains the sender's ECDH public key, so the receiver would be able to recover the shared nonce using its `blinding_private_key`.
 
 An additional use-case is sharing some/all of the receiver's blinding private keys with an external auditor, allowing unblinding the audited outputs without being able to spend them.
 
--- a/slip-0077.md
+++ b/slip-0077.md
@@ -1,0 +1,62 @@
+# SLIP-0077 : Deterministic blinding key derivation for Confidential Transactions
+
+```
+Number:  SLIP-0077
+Title:   Deterministic blinding key derivation for Confidential Transactions
+Type:    Standard
+Status:  Draft
+Authors: Roman Zeyde <mail@romanzey.de>
+Created: 2019-06-15
+```
+
+## Abstract
+
+This document describes a method for blinding key derivation
+for Confidential Transactions, using a determinstic hierarchy.
+
+## General design
+
+In confidential transactions, the sender and the receiver use ECDH to derive a shared nonce, which is then used for hiding/recovering of the actual value and asset type being transacted.
+In Elements/Liquid, the receiver uses the following derivation scheme for his ECDH public/private keys:
+
+```
+blinding_private_key := HMAC_SHA256(key=master_blinding_key, msg=script_pubkey)
+blinding_public_key := secp256k1_publickey(private_key=blinding_private_key)
+```
+
+The receiver is using `blinding_public_key` construct a "blinded address", which is used by the sender to blind the relevant transaction outputs. Each such blinded transaction output also contains the sender's ECDH public key, so the receiver would be able to recover the shared nonce using its `blinding_private_key`.
+
+An additional use-case is sharing some/all of the receiver's blinding private keys with an external auditor, allowing unblinding the audited outputs without being able to spend them.
+
+## Design details
+
+### Master blinding key derivation
+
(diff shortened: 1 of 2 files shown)


romanz:
Updated the SLIP following https://github.com/trezor/trezor-firmware/issues/244#issuecomment-504119392.

slip-0077.md
```diff
@@ -0,0 +1,62 @@
...
+
+The receiver is using `blinding_public_key` construct a "blinded address", which is used by the sender to blind the relevant transaction outputs. Each such blinded transaction output also contains the sender's ECDH public key, so the receiver would be able to recover the shared nonce using its `blinding_private_key`.
+
+An additional use-case is sharing some/all of the receiver's blinding private keys with an external auditor, allowing unblinding the audited outputs without being able to spend them.
+
+## Design details
+
+### Master blinding key derivation
+
+In order to use similar blinding key derivation scheme on TREZOR, we suggest using the following [derivation scheme](https://docs.google.com/document/d/1QtwF8DF0TyPt9KFp5EsVpOY5d8PMicqafb2yGFoKfZ0/) for `master_blinding_key`:
```
andrewkozlik: Reference SLIP-0021 once it gets approved.

slip-0077.md
```diff
@@ -0,0 +1,62 @@
...
+
+```
+domain := b"Symmetric key seed"
+root   := HMAC_SHA512(key=domain, msg=seed)
+
+label := b"SLIP-77"
+node  := HMAC_SHA512(key=root[0:32], msg=label)
```
andrewkozlik: Assuming SLIP-0021 gets approved in the current form, add a null byte before the label, i.e. msg = b"\x00" + label.

slip-0077.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Master blinding key derivation
+
+In order to use similar blinding key derivation scheme on TREZOR, we suggest using the following [derivation scheme](https://docs.google.com/document/d/1QtwF8DF0TyPt9KFp5EsVpOY5d8PMicqafb2yGFoKfZ0/) for `master_blinding_key`:
+
+```
+domain := b"Symmetric key seed"
+root   := HMAC_SHA512(key=domain, msg=seed)
+
+label := b"SLIP-77"
```
andrewkozlik: Preferably use "SLIP-0077" for consistency.

prusnak:
JFYI https://github.com/satoshilabs/slips/blob/master/slip-0021.md has been just merged

romanz:
Many thanks, will update the PR.

romanz:
Squashed and force-pushed 2992f3f.

andrewkozlik:
> Just one note. In the very unlikely event (< 2^-127) that `blinding_private_key` is bigger than or equal to the curve order, `secp256k1_publickey` fails, since it assumes the private key is less than the order.

I see the following options:

1. Leave it as is. Mathematically the operation blinding_public_key = blinding_private_key * G is well defined. The fact that trezor-crypto requires the private key to be in a certain range is merely a restriction of the implementation. The programmer implementing SLIP-0077 with trezor-crypto should use blinding_private_key mod secp256k1_curve_order to comply with this restriction.
2. Define blinding_private_key := HMAC_SHA256(key=master_blinding_key, msg=script_pubkey) mod secp256k1_curve_order. This is equivalent to the previous option, but ensures that nobody makes a mistake implementing it.
3. Define blinding_private_key := HMAC_SHA256(key=master_blinding_key, msg=script_pubkey) mod 2^255. This is not equivalent to either of the above. It loses one bit of entropy, but ensures that all possible blinding keys have uniform distribution of probability.

In cases 1 and 2 the distribution of the keys is non-uniform, which is something I was a little concerned about when I first saw your proposal. However, realizing that the likelyhood of hitting one of the keys which have higher probability is about 2^-127, I wasn't bothered about it. If you are already using this scheme in production, then 3 isn't an option. In that case I would see 2 as the best solution just to be on the safe side, but I am fine with any of the above.

slip-0077.md
```diff
@@ -0,0 +1,62 @@
...
+
+In confidential transactions, the sender and the receiver use ECDH to derive a shared nonce, which is then used for hiding/recovering of the actual value and asset type being transacted.
+In Elements/Liquid, the receiver uses the following derivation scheme for his ECDH public/private keys:
+
+```
+blinding_private_key := HMAC_SHA256(key=master_blinding_key, msg=script_pubkey)
+blinding_public_key := secp256k1_publickey(private_key=blinding_private_key)
+```
+
+The receiver is using `blinding_public_key` construct a "blinded address", which is used by the sender to blind the relevant transaction outputs. Each such blinded transaction output also contains the sender's ECDH public key, so the receiver would be able to recover the shared nonce using its `blinding_private_key`.
```
real-or-random: nits:
 * typo: "to" construct
 * hm, "blinded address" is a somewhat misleading term because the address itself is not blinded

onvej-sl:
> Mathematically the operation blinding_public_key = blinding_private_key * G is well defined. The fact that trezor-crypto requires the private key to be in a certain range is merely a restriction of the implementation.

The operation `blinding_public_key = blinding_private_key * G` is well defined, but the standard refers to the function `secp256k1_publickey`, which is defined [here](https://github.com/trezor/trezor-firmware/blob/5aa592a69af8bdf14cb98f01941174b82c753b34/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1.h#L61).

onvej-sl:
> IIUC, Elements' code returns an "invalid" key if the private key is bigger than the group order:
https://github.com/ElementsProject/elements/blob/7b302eb99d2d1bf861429d6bd209771f634b8e8e/src/wallet/wallet.cpp#L5583-L5617

Well done. Maybe it should be written in the standard, just to show we are aware of the possibility of 'overflow'.

slip-0077.md
romanz: Thanks!
Should be fixed at e394776.

romanz:
> Maybe it should be written in the standard, just to show we are aware of the possibility of 'overflow'.

Thanks!
Added a note about this issue at e394776.

PR #663, opened 2019-06-17
Add YCash(YEC)

denverbdr:
Add CoinCode for Ycash(YEC).


prusnak:
choose the first available number

denverbdr:
Fixed, thank you for the guidance. 

Picked the first free number

PR #672, opened 2019-06-26
Add Aeon currency

BigslimVdub:
Add Aeon currency to the SLIP-0044 document. Website may change to https in the near future but will redirect to https if http is used after the change.


prusnak:
please use the first available number, also submit when you have a working https

BigslimVdub:
Ok, I think that is good now. Sorry I forgot to update from other merged coins.

prusnak:
Remove the trailing `|` character. Use number 355. The 354 got allocated in the meantime

PR #680, opened 2019-07-03
Added ABBC constant

abbc-foundation:
New constant for ABBC coin (https://www.abbcfoundation.com/)


prusnak:

abbc-foundation:
Can we use 0xABBC constant? Similar to IOTA using 0x107A?

prusnak:
No

abbc-foundation:
Ok, understood, is 0x800001ab available?

prusnak:

abbc-foundation:
Updated to use the first available constant

PR #693, opened 2019-07-10
Add cruzbit

GreenAppers:


prusnak:

asdvxgxasjab:

We're already using this number. There's already tons of sparse entries in the list. What's the problem?

prusnak:
the problem is you have started using the number before allocating it first

GreenAppers:
Right. We already launched using this number.  We have users with substantial cruzbit holdings in HD wallets generated using 831 as the coin type. We can't change at this point.

I've updated our documentation to refer to https://github.com/GreenAppers/slips and we will continue to maintain the fork.

For future reference the criteria of "use the first available number" should be documented more prominently.  SLIP-0044  is filled with counter examples and there is no such indication that I've seen before your comment.

PR #702, opened 2019-07-21
Adding Xerom to slip-0044.md

fallengravity:
Name: Xerom
Ticker: XERO
Explorer: https://explorer.xerom.org
Public RPC: https://rpc.xerom.org

Xerom is an Ethereum Fork with a modified Monetary policy, it also includes a new Masternode system which is run on-chain.

CC: @mkrufky


prusnak:

mkrufky:
@prusnak , for over a year now, I have been encouraging new chains to pick a SLIP0044 offset that matches the chain ID exactly.  Please approve this using the number `1313500` as-is.

PR #704, opened 2019-07-22
SLIP-0077: Correct ECDH pseudo-code

stevenroose:
`secp256k1_multiply` doesn't exist. `secp256k1_ecdh` is used to do ECDH and there is no compression flag present: https://github.com/bitcoin-core/secp256k1/blob/c8fbc3c397b547bc64435a9bffb8f989cd23aba0/include/secp256k1_ecdh.h#L42

### DIFF
--- a/slip-0077.md
+++ b/slip-0077.md
@@ -51,7 +51,7 @@ master_blinding_key := node[32:64]
 The shared nonce is derived using ECDH and double-SHA256 of the compressed shared public key:
 
 ```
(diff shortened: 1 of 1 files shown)


stevenroose:
Hmm, it seems that `secp256k1_multiply` confused me with `secp256k1_ec_pubkey_tweak_mul`. I think `secp256k1_multiply` is just used as a "do an EC multiplication" pseudo-method.

PR #706, opened 2019-07-23
Add coin type SEOS by SuperEOS

FlavoredYoghurt:
Add coin type SEOS by SuperEOS


prusnak:

Issue #703, opened 2019-07-22
SLIP-0010: Add BIP32-ED25519 Curve


ilap:
The ed25519 curve lacks of the public child key generation which is required, for example, for developing a monitoring only (watch-only) wallet. See, a dart code example below.

However, there is a BIP32 compliant version of the ed25519 ([BIP32-ED25519](https://cardanolaunch.com/assets/Ed25519_BIP.pdf)), which is backward compatible /w the original ed25519 (all the signing and verifying procedures remain the same), that produces all its extended private keys that lie in some affine space and therefore it can have a public child key generation function that comply /w the BIP32 specification, and therefore should be added to the list of the supported curves.

``` dart
var mStr =    'xprv9s21ZrQH143K3gks4d1bQXQ3RUJ8B2DtJepewENRru1GeQR1Dzwn4wwTZSGP22aSyyuMKVWjQyMQMmSHeeHS69rjcDySQyJLVAw2XFv4xXi';
var m = Chain.import(mStr);

var xprv = m.forPath("m/44'/0'/0'");
var xpub = xprv.publicKey();
var address = m.forPath("m/44'/0'/0'/0/0");

var watchOnly = Chain.import(xpub.toString());
var watchAddress = watchOnly.forPath("M/0/0");

print("======= Addressess are the same =======");
print("address (m/44'/0'/0'/0/0): ${address.publicKey().toString()}");
print("watchAddress (M/0/0):      ${watchAddress.toString()}");
```

andrewkozlik:
As you have already noticed, I created SLIP-0023 for this purpose. I had three reasons to create a separate spec for this:

1. The "Ed25519-BIP32" scheme is similar, but not compatible with SLIP-0010 or BIP-0032. For one thing, it uses a different child key derivation method.

2. There is a lot of confusion about the derivation of the master node in "Ed25519-BIP32". There is one method which is described in the original paper, but which is not implemented anywhere. Then I found that there are three methods which are implemented, two of which are used in practice (Daedalus and Icarus), but none of these is documented anywhere. All four methods are mutually incompatible and none them uses HMAC-SHA512 of the seed with the curve name as is done in BIP-0032 and SLIP-0010. This is important to ensure proper domain separation between different elliptic curves or different types of key hierarchies. I would rather not bring all this confusion and malpractice into SLIP-0010.

[...]

PR #712, opened 2019-07-24
slip-0023: Create SLIP-0023.

andrewkozlik:
This specification describes a new method of deriving the master node of the key tree in Cardano hierarchical deterministic wallets which is primarily intended for use with SLIP-0039 Shamir's Secret-Sharing for Mnemonic Codes.

### DIFF
--- a/slip-0023.md
+++ b/slip-0023.md
@@ -19,7 +19,7 @@ Cryptocurrency wallets generally work by combining a [BIP-0039](https://github.c
 
 Cardano hierarchical deterministic wallets use an extended private key which requires them to implement a custom adaptation of the BIP-0032 child key derivation scheme. The Cardano wallets currently in existence implement several mutually incompatible schemes for the derivation of the master node from a BIP-0039 mnemonic and passphrase. Unfortunately all of these derivation schemes fail to separate the derivation of the seed from the derivation of the key tree, making it impossible to integrate them with new seed derivation schemes. This specification aims to rectify this by defining a new scheme for the derivation of the master node from a seed. The new scheme is primarily intended for use with SLIP-0039 Shamir's Secret-Sharing for Mnemonic Codes.
 
-## Universal master node derivation
+## Cardano universal master node derivation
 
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
(copying what I posted on the forum here)

Just an FYI that both the v1 (Daedalus) scheme and the v2 (Icarus) scheme will be deprecated for Shelley in favor of a v3 scheme that uses bech32. The Rust team initiated the migration and the Haskell team wrote a bech32 library for Haskell to match it https://input-output-hk.github.io/cardano-wallet/haddock/edge/bech32-2019.5.8/Codec-Binary-Bech32.html 1

SebastienGllmt:
I found the original spec by the Rust team for bech32 which my be useful to you
https://github.com/input-output-hk/implementation-decisions/blob/master/text/0001-address.md

ilap:
It should be integrated into the [SLIP-10](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) by more generalising the SLIP-10's master key generation and key derivation functions.
I ahve already generated an [issue](https://github.com/satoshilabs/slips/issues/703) a couple of days ago.

andrewkozlik:
> I found the original spec by the Rust team for bech32 which my be useful to you
> https://github.com/input-output-hk/implementation-decisions/blob/master/text/0001-address.md

I seem to be missing something, because as far as I am able to tell this spec is about encoding the addresses. It should not impact the derivation of the master node.

ilap:
Yes, that's only for the address generation.
But, I could not find the validation part of the generated master key, the 3rd bit of the _k[31]_ must be 0.

ilap:
Oh, you clear all its highest 3 bits, /w the _k[31] & 0x1f_.

sterlingwellscaffeine:
Dear Team, I am new to this project but I am curious about the second step in the algorithm for deriving the master node: 

```Determine the initial entropy *E* that was used to generate *M*.```

How is this done? I assumed that the RNG that uses E to generate M would be a 1 way function from whence it would be impossible to recover E from M. Is this not the case? 

Thank you so much in advance for your advice.

andrewkozlik:
> Dear Team, I am new to this project but I am curious about the second step in the algorithm for deriving the master node:
> 
> `Determine the initial entropy *E* that was used to generate *M*.`
> 
> How is this done? I assumed that the RNG that uses E to generate M would be a 1 way function from whence it would be impossible to recover E from M. Is this not the case?

BIP-39 starts with initial entropy *E* converts that to a mnemonic *M* and from the mnemonic (+ optional passphrase) you get a seed *S*. The conversion from *E* to *M* is easy, just map the words to 11-bit numbers and concatenate the 11-bit chunks https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic. Converting from *M* to *S* uses a one-way function, so going from *S* to *M* would be computationally infeasible.

sterlingwellscaffeine:
That was my understanding as well. I think my misunderstanding was I have been working more with the generated seeds than I have with the mnemonic in general so my reading of the algorithm thought that I had to go from S -> E. If the algorithm as stated in the document requires going from M -> E, that is doable. Thank you for the explainer!

PR #715, opened 2019-08-01
slip-0044: Add new coin KAL

kellbenet:
Add new coin for the Kaleidochain. KAL for mainnet.

Our codebase is [Kaleidochain](https://github.com/kaleidochain/kaleido) at GitHub. We implement a new standalone blockchain using newly algorand pure PoS consensus algorithom.

Thanks for your time.


prusnak:
Testnets should always use `1` as bip44 value. Please update your commit

kellbenet:
Thanks @prusnak. You are right. I have changed to use the first available value only for mainnet.

PR #718, opened 2019-08-03
slip-0044: add NEAR

potatodepaulo:
Add NEAR coin to registry.


prusnak:

potatodepaulo:
Why are other numbers added at much higher increments? I don't understand this policy.

prusnak:
Unless you have a good reason just follow the policy of using the lowest available number

vgrichina:
@prusnak used lowest available number

PR #719, opened 2019-08-05
Fix ICON's coin type value

cwkwak:
ICON has used 0x80494358 for the coin type. (applied to ledger app)


prusnak:
Fix the ledger app to use the correct number.

cwkwak:
We released the ledger app about 1 year ago. Many users have already created and used keys. Could you accept my pull request?

prusnak:
Why haven't you used the assigned number?

prusnak:
Also, no, I am not changing the assigned number. There is no way how to determine how many applications are there which use the correct number .

cwkwak:
1 year ago, it was not defined. One of our team members was supposed to register this number, but he missed. Recently I found out and submitted the pull request.
I’ll check with the person to register 4a where he’s used it.

PR #723, opened 2019-08-07
Added SIN to the list

giaki3003:


prusnak:

giaki3003:
Hello, we already have some HD wallets from previous core implementations using this derivation path, changing it would be quite difficult and give users some headache; if possible we'd like to maintain this for the list even if it's not the first immediate available option. Thanks for understanding

PR #724, opened 2019-08-08
slip-0022: Create SLIP-0022.

andrewkozlik:
@onvej-sl please check the examples.


andrewkozlik:
> Examples are OK, but there a few things that were a little confusing for me:
> 
> > A CredentialData DOM object defined as follows:
> 
> Why is the object called "DOM" object?
> 
> > `3: h'3082019330820138A0030201023082019330820138A003020102308201933082',`
> 
> What is the meaning of the "h" prefix?

I just reused the terminology and notation from the CTAP spec without giving it much thought, but I agree it would be better to be more precise here and talk about a "credential data object in CBOR diagnostic notation". See RFC 7049.

PR #734, opened 2019-08-15
Added PKT chain to slip-0044 with next available number (390)

cjdelisle:
Title pretty much says it all, please let me know if there are any issues.


cjdelisle:
the ugly force-push is because I started making another PR using the same branch and when I pushed it, it updated this one so I reset back by 1 and forced it.

PR #742, opened 2019-08-20
slip-0044: add CRO

tomtau:
393 (0x80000189) was the first available number at the time of PR submission


prusnak:
Please use 394, 393 was allocated in PR #740

tomtau:
Changed to 394 (0x8000018a)

PR #749, opened 2019-08-27
Add KEX to sip-0044.md

asmodat:
Please add [Kira Exchange Token](https://kiraex.com/) to sip-0044.md


prusnak:

asmodat:
We wanted to do so but 1026 decimal is a Gematria for the "Kira Exchange Token", could we keep it ?

PR #757, opened 2019-09-07
Add Galilel (GALI) 402 to slip-0044

mbroemme:
Added [Galilel (GALI)](https://galilel.org/) coin to slip-0044.md


prusnak:

mbroemme:
@prusnak Now using first available number 402 and squashed the commits into single one.

PR #779, opened 2019-09-20
add Rust implementation of SLIP-0039

tamasblummer:
You might want to add a pointer to this implementation in Rust

### DIFF
--- a/slip-0039.md
+++ b/slip-0039.md
@@ -315,6 +315,10 @@ JavaScript:
 
 * <https://github.com/ilap/slip39-js>
 
+Rust:
+
+* <https://github.com/rust-bitcoin/rust-wallet/blob/master/src/sss.rs>
+
 ## Design rationale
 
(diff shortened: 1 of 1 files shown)


prusnak:
I think you should rename the class and filename to Slip39 if this is an implementation of this particular standard. Otherwise OK.

tamasblummer:
I rather hope that you set the standard for the use of this scheme therefore not referring to by slip but by functionality. Similar to how the library refers to Wallets and not BIP32

prusnak:
If you don't plan to change the scheme in your library then I am OK with it. Merged.

tamasblummer:
That library has also other TREZOR compatible code. You find BIP44, BIP48, BIP84 implementations next to this. I promise not to move the linked file.

tamasblummer:
and BIP39 of course

prusnak:
Yeah, I checked your project, really nice!

PR #782, opened 2019-09-25
slip-0044: add CET

whunmr:
Add CET.

- site: https://www.coinex.org/
- testnet: 
    - https://testnet.coinex.org/
    - https://github.com/coinexchain/testnets/

- wallets:
    - https://wallet.viabtc.com/?lang=en_US
    - https://www.purewallet.org/
    - https://www.ifwallet.com/


prusnak:

whunmr:
OK. got it.

whunmr:
Since our community developed the chain and wallets since several months ago.
would be good if we can use the 688. 

IF 'use the first available number' is mandatory,  we are also OK to change the PR soon.

prusnak:
If you already use 688 we can use that.

whunmr:
Ok, Then I'll change the PR back to 688.   Thanks @prusnak

PR #783, opened 2019-09-25
slip-0044: add evan.network to SLIP List

S3bb1:
Add EVE.

website: https://www.evan.network
wiki: https://evannetwork.github.io/
github: https://github.com/evannetwork

testnet:

website: https://dashboard.test.evan.network
explorer: https://testexplorer.evan.network
spec: https://github.com/evannetwork/testcore-config

mainnet:

website: https://dashboard.evan.network
explorer: https://explorer.evan.network
spec: https://github.com/evannetwork/core-config


prusnak:

S3bb1:
Ah okay, i am a little bit confused because some coins are using their chainid as the number. so the next free number should be used instead?

prusnak:
Ah, it was not obvious this is Ethereum fork. You can decide whether you want to have chain_id or next free number.

S3bb1:
then we switch back ot the chain id :) thank you for clarification!

PR #786, opened 2019-10-01
Change to TON web address

Futurizt:
the official TON project address is https://test.ton.org
what is listed now is a scam website. Will report to github if not changed.


prusnak:
Fix the url, it is not in the proper format now

PR #793, opened 2019-10-11
Add ZCore

apollomatheus:


prusnak:

apollomatheus:
Thanks @prusnak, all done.

PR #798, opened 2019-10-15
add WE coin

maulikvora:


prusnak:
use https

maulikvora:
Did it. Thanks for quick response

PR #803, opened 2019-10-23
slip-0044: add YEE(YeeCo)

guoxbin:


prusnak:

guoxbin:

I see there are many coins use meaningful or optional coin type numbers, such as :
ONT: 1024
ACT: 666
BNB: 714 (Binance was founded on 7.14)

I choose a omitted number 4096 which is between 2 existing number ( 3377 and 4218) and will not expand the number range in use.
I'd appreciated you consideration.

@prusnak

guoxbin:

any response?
@prusnak

prusnak:

guoxbin:

Thanks for your response. 
We prefer a meaningful number as some other coins  and we humbly request your consideration.
Why we prefer 4096. that's 4 * 2^10. 
Our blockchain project is related with 4k video (high definition video) transmission. 
One of the projects in our ecosystem is YeeCall (http://www.yeecall.com) which is dedicated in high definition VOIP, 4k video conference system. 
When we find 4096 has not been used (between 2 used number, 3377 and 4218), we were so excited. 
Thanks for your time.

PR #813, opened 2019-11-05
0173: Add IoTeX prefix ('io').

catenocrypt:
Defined here:  https://github.com/iotexproject/iotex-address/blob/master/address/address.go


prusnak:
use url for the project, not github repo

catenocrypt:
> use url for the project, not github repo

Changed as suggested.
Note that I have included a reference to a file where the prefix in question is defined (https://github.com/iotexproject/iotex-address)

PR #816, opened 2019-11-12
slip-0022: Add algorithm field.

andrewkozlik:


onvej-sl:
I'm not sure whether I understand what is the purpose of the `curve` field. Is its purpose to distinguish between Ed25519 and Ed448 if the algorithm is EdDSA?

If we are planning to use another algorithm that needs parameters in future we might want create a nested structure called `algorithmParameters`.

andrewkozlik:
> I'm not sure whether I understand what is the purpose of the `curve` field. Is its purpose to distinguish between Ed25519 and Ed448 if the algorithm is EdDSA?

Yes.
 
> If we are planning to use another algorithm that needs parameters in future we might want create a nested structure called `algorithmParameters`.

As discussed in meeting, we want to keep the structure flat.

slip-0022.md
```diff
@@ -186,6 +192,7 @@ The rationale behind using the authentication tag for the SLIP-0010 key path is
 
 * [Web Authentication](https://www.w3.org/TR/webauthn/): An API for accessing Public Key Credentials Level 1, W3C Recommendation, 4 March 2019
 * [FIDO Client to Authenticator Protocol (CTAP) v2.0](https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#sctn-hmac-secret-extension), Proposed Standard, January 30, 2019
+* [IANA-COSE-REG](https://www.iana.org/assignments/cose/cose.xhtml#algorithms): IANA CBOR Object Signing and Encryption (COSE) Registries.
```
onvej-sl: Should be without `#algorithms` at the end of the url.

PR #822, opened 2019-11-19
Add COCOS to SLIP-0044

SYLing:
Add COCOS


prusnak:

SYLing:
Add COCOS

prusnak:
the first available number is 446

PR #826, opened 2019-11-25
Added coin type for Ethereum 2

mcdee:
Ethereum 2 continues to use Ether as its network currency, however it is stored and addressed in a different fashion.  Most notably, the curve used is BLS12-381.  As such, although they are technically the same coin they are different coin types and a new ID is required to avoid potential issues.

This allocates a new coin type for Ethereum 2: 3600.  This value has been listed in ERC-2334 and will become the standard for all HD key derivation systems for this chain.


prusnak:
Citing from the EIP: " This new scheme is incompatible with the current form of this specification BIP44 ...", so it really does not make sense to allocate the number here.

PR #829, opened 2019-12-04
slip-0044: add Energi (NRG) Gen 3

andvgal:


prusnak:
use first available number, testnets should always use `1` as slip44 value

andvgal:
Please clarify where the ruling comes from:

1. There is no BIP44 notion of all testnets using the same path while registration of testnet is by nature of the spec.

**Justification**: it is essential to uncover all places which may use the "default" HD paths instead of required for the project during integration testing in development which must be isolation from production according to best practices.

2. There are clear rules:

> ... Cryptocoin developers may ask for registering unused number for their project. 

That what we exactly do - we have a new project done from scratch and a new registration, accompanied by testnet reg.

> To register a new coin type, an existing wallet that implements the standard is required and a pull request to the above file should be created. 

This requirement is fulfilled.

prusnak:
SLIP-0044 says "index:1 - Testnet (all coins)"

PR #836, opened 2019-12-15
Add Capricoin+

CapricoinPlus:
This is the new coin (different chain) that is meant to be a swap for the old unmaintained Capricoin


prusnak:
Why don't you use the same number?

CapricoinPlus:
to avoid confusion, the new source doesn't resemble anything to the old one.

PR #843, opened 2019-12-25
[New SLIP] A Standard for Universal Multi-currency Addresses

ty-everett:
I'd like to publish a new SLIP that defines a standard for a universally-accepted cryptocurrency address format. The standard isn't too rigid and maybe people won't adopt it, but I think the benefits in trying to establish something like this are far too great for me not to make an attempt.

Here are some ways that this might substantively benefit the cryptocurrency ecosystem:

- Improve UX on exchanges and wallets that deal with multiple currencies
- Give each project a pool of addresses to work with
- Increase transactional privacy for all projects (because "Enhanced Privacy Mode" makes it impossible to correlate two separate UTXOs as being part of the same payment)

I originally tried to publish this as a BIP, but the BIP editor said that it was out of scope. I saw the README for this repository and so I hope this standard can find a home here.

Please let me know if you have any questions. You can also check [universaladdresses.com](https://universaladdresses.com) for more details. A list of participating organizations will be added to the site once a final BIP/SLIP number can be assigned. For now I left it at 3301.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,30 +11,31 @@ and contains the documents that are unsuitable for submission to BIP repository.
 
 Each SLIP should provide a concise technical specification of the feature and a rationale for the feature.
 
-| Number                    | Title                                                                 | Type          | Status   |
-|---------------------------|-----------------------------------------------------------------------|---------------|----------|
-| [SLIP-0000](slip-0000.md) | SLIP Template                                                         | Informational | Accepted |
-| [SLIP-0010](slip-0010.md) | Universal private key derivation from master private key              | Standard      | Draft    |
-| [SLIP-0011](slip-0011.md) | Symmetric encryption of key-value pairs using deterministic hierarchy | Standard      | Draft    |
-| [SLIP-0012](slip-0012.md) | Public key encryption using deterministic hierarchy                   | Standard      | Draft    |
-| [SLIP-0013](slip-0013.md) | Authentication using deterministic hierarchy                          | Standard      | Draft    |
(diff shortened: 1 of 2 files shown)


prusnak:
I think in order for the standard to become an Universal Addresses you need to address the inability to use different curve, i.e. this standard is limited to secp256k1. Maybe you want to use SLIP-10 instead of BIP-32?

ty-everett:
That is a valid concern. At the end I talk about the inability to use Ed25519 because public derivation is not supported. In SLIP-0010 also public derivation is not supported on Ed25519.

The reason BIP-0032 was used instead of using SLIP-0010 is also because:

- Serialization format for BIP-0032 is well-recognized and used by the community, including many wallet implementations.
- For cryptocurrency addresses, NIST P-256 is not a common curve because normally secp256k1 is used
- secp256k1 is much faster to perform operations on compared to P-256
- If a secp256k1 address format was used by some large wallets and exchanges then other projects that used Ed25519 (XMR) may decide to switch over to secp256k1 later.
- The primary objective is to build a standard that is compatible with a lot of existing implementations. Regardless, since public derivation for Ed25519 is not solved, sacrificing the universality of BIP32 serialization would have only a very small benefit.
- For SSH and GPG, it is posible to use and configure those software to use secp256k1 as well. And when that configuration is made then their cryptographic operations will also be much faster than P-256 :)

prusnak:
If you are omitting p-256 and ed25519 curves then your standard is hardly universal, but rather quite limited. It's not worth adding it to our SLIPs.

PR #849, opened 2020-01-02
slip-0044: Add UND

Codegnosis:
Wallet SDK: https://github.com/unification-com/und-js


prusnak:

PR #853, opened 2020-01-07
slip-0044: add DxChain Token [DX]

mzhangdx:
Added DxChain Token [DX] to slip-0044.md


prusnak:
use the first available number; left align the symbol

mzhangdx:
> use the first available number; left align the symbol

1. Switched the number to 467, which is the first available number in the document
2. Left aligned the symbol

Thank you.

PR #860, opened 2020-01-17
slip-0044 add Origo [OGO]

yijia-origo:


prusnak:

yijia-origo:

Fixed

PR #877, opened 2020-02-17
add BNT to slip-0044

njmurarka:

PR #878, opened 2020-02-18
slip-0044: add AMO

Lbird:
Use the first available number 484 after PR #877 .


prusnak:
Fixed in 7362e452576df3e757181e522641b781d264a1df

PR #886, opened 2020-02-27
Added entry for Amplifi (2874)

tohsnoom:


prusnak:
use the first available number, use URL for coin

prusnak:
No response.

PR #887, opened 2020-02-27
Added coin type Symbol (XYM)

evias:
## Added

- Added coin type for `Symbol` with ticker `XYM`
- Link: https://github.com/nemtech/catapult-server


prusnak:

evias:
@prusnak thank you for the fast response ; is it really not possible to use `4343` ? 

The reason why we asked to use that one is because it "relates to NEM", as Symbol is our next blockchain (Symbol from NEM).

I would appreciate if it's possible to use `4343` after all, seeing that a lot of projects have been given custom numbers.

prusnak:
Merged. I encourage you to donate some Bitcoins here: https://trezor.io/beer/

evias:
Thank you for merging :+1: enjoy the beer(s) :)

PR #890, opened 2020-02-28
Add Counos Coin. CounosX and Counos E coins

counos:


prusnak:
I suggest using the same value for all your Counous coins. Ethereum tokens also don't have different paths for every token. Also use the first available number.

counos:
Counos has different coins with different network, they dont have same chain and same structure.
I see that there are some free space, if I can use index 489, and  I will use it

prusnak:
Use 489

prusnak:
I am not allocating 20 numbers for 20 of your coins. Please make just one allocation - for 489.

prusnak:
added in 0dc49bd1e199fcf401f41894957e63026b692694

PR #893, opened 2020-03-04
Update slip-0044.md

finlchain:
Add FIN


prusnak:
use lower case letters for hex numbers

finlchain:
I changed to lower case letters for hex numbers.
What is the meaning of the first available number? Do you say number 493?

prusnak:
Yes, please use 493

finlchain:
Changed to the number 493.

PR #896, opened 2020-03-06
Add BHT to slips-0044

kai-bhex:


prusnak:

kai-bhex:
Change to 496, as 495 is in pending pr

PR #901, opened 2020-03-10
Add Totem Live Accounting XTX to slip-044

chrisdcosta:
Totem Live Accounting is a Substrate based blockchain which already has a live public network.

* We have linked to the live network which contains the wallet implementation (Identities) in the SLIP request.

* Identities are BIP-0039 compatible and can be restored from a seed phrase + optional derivation paths.

* Restored Identities are also compatible with Parity Polkadot (DOT) and Kusama (KSM) networks

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -1123,7 +1123,7 @@ index | hexa       | symbol | coin
 13107 | 0x80003333 | BTY    | [BitYuan](https://www.bityuan.com)
 13108 | 0x80003334 | YCC    | [Yuan Chain Coin](https://www.yuan.org)
 15845 | 0x80003de5 | SDGO   | [SanDeGo](http://www.sandego.net)
-16181 | 0x80003F35 | XTX    | [Totem Live Network](https://totem.live)
+16181 | 0x80003f35 | XTX    | [Totem Live Network](https://totem.live)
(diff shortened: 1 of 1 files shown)


prusnak:

chrisdcosta:
We have already deployed our chain, is there no flexibility?

prusnak:
I don't see how the chain is related to SLIP-44 allocation.

prusnak:
Or do you want to use the same slip44 number as the chain_id?

chrisdcosta:
Yes, that's the idea. We are following Parity's [Subkey](https://substrate.dev/docs/en/ecosystem/subkey#hd-derivation)  / polkadot-js implementation which also allows for a variety of extended derivation path combinations (including strings) - so it goes beyond the Bitcoin standard. 

Parity has distinguished between address generation on the Polkadot Network `354   | 0x80000162 | DOT` and their Kusama Network `434   | 0x800001b2 | KSM ` otherwise the addresses would be identical for both chains, and as we are also using the same address generation methods - it would even be identical for Totem's chain - unless we make a distinction. 

We have used this number `16181` (a variation on the golden ratio) in several other contexts as a pseudo chain Identity, so it would be nice if we could also have it here.

PR #902, opened 2020-03-12
slip-0044: add CS

0xAAE:


prusnak:

0xAAE:
I've used the 1007 in pull request that is available, let me know if it's ok

prusnak:
it's not that why i asked you to use the first available number

PR #910, opened 2020-03-30
Apply for the TTC updating of slip-0044

TTCOffical:
We are developing the ledger app of TTC, so we would like to register the new coin type for TTC
Thanks


prusnak:

prusnak:
No reaction. Closing

PR #912, opened 2020-04-02
Apply for the GateChain updating of slip-0044 

vincentbdb:
The GateChain is going finish developing, so we want to register the new coin type for GC(GateCoin) :669
Thanks


prusnak:
Remove `?lng=en` from the url

vincentbdb:
> Remove `?lng=en` from the url

Done
Thanks

PR #918, opened 2020-04-21
SLIP-0044: Add entry for Elrond

andreibancioiu:
We are currently configuring key derivation for our wallet & wallet-like applications and for our extension to TrustWallet Core library - as seen below:

https://github.com/ElrondNetwork/wallet-core/blob/new-coin-elrond/coins.json#L1340

Therefore, we would like to add an entry for ERD in the SLIP-0044 registry.


prusnak:
What's the chain_id for this coin?

iulianpascalau:
We haven't decide for a mainnet chain ID yet as we are in the final testing sessions rounds. 
If it is a must, we will decide upon it ASAP.

prusnak:
If you don't have a chain_id then please use the next available number, not a random one.

Another option is that you use the same chain_id as suggested here for slip44 as having the same chain_id and slip44 number makes sense.

andreibancioiu:
Thank you @prusnak. We've updated the PR to use the first available number.

PR #920, opened 2020-04-24
Add Newton coin type.

arisac:
add NEW


prusnak:
use the first available  number

arisac:
What is the meaning of the first available number? 

We are using the number 1642.

prusnak:
You should register the number first, then start to use it. We assign numbers sequentially.

If you already use this number by accident, show me the code where it is used.

arisac:
Our project is already using the number 1642.

Here is the example code to our project： https://github.com/newtondevelop/newpay-android-example/blob/master/src/app/src/main/java/org/newtonproject/newtoncore/android/C.java#L55

I apologize that we didn't make a formal registration before.

And we still what to use the number 1642 for our project. 

Thank you.

prusnak:
1642 is not 0x658 but 0x66a

PR #923, opened 2020-05-05
Added XDC Blockchain

AnilChinchawale:


prusnak:

AnilChinchawale:
We have ChainID :- 50 hence use most available number i.e. 550

PR #928, opened 2020-05-11
add entry for two stablecoins: chf36 and eur36 on ethereum mainnet

wstrametz:
Hi - hope I understood everything properly - I want to register two new stablecoins for euro and swiss francs - github repo: https://github.com/element36-io.  Thx


prusnak:
Ethereum tokens should not use custom paths but stick to Ethereum paths.

PR #936, opened 2020-05-21
SLIP-0019

andrewkozlik:
For design details see https://docs.google.com/document/d/1nvMxTGFcpIUvSxcuHsyalJ0GwZYCuTbB8yso49u39j0/

@stepansnigirev feel free to comment. I am not sure why, but I couldn't add you as a reviewer.

### DIFF
--- a/slip-0019.md
+++ b/slip-0019.md
@@ -141,6 +141,8 @@ Each input which is registered to take part in a CoinJoin transaction should be 
 3. Verify that bit 0 (user confirmation) of *flags* is set.
 4. Verify that the *proofSignature* is valid using the obtained scriptPubKey.
 
+A proof of ownership commits to a particular scriptPubKey, which means that the proof is replayable for UTXOs with the same address. Nevertheless, freshness of such a proof is guaranteed if a nonce (such as the *psbtId*) is included in the *commitmentData*.
+
 ## PSBT (BIP 174) extension
 
 The following new global field type is added to the BIP-0174 specification:
--- a/slip-0019.md
+++ b/slip-0019.md
@@ -11,63 +11,156 @@ Created: 2019-04-25
 
 ## Abstract
 
-This specification defines the format for a Proof of Ownership which can be passed to a hardware wallet together with an unsigned transaction. This proof allows the hardware wallet to determine whether the input is controlled by the key owned by the hardware wallet or not. It also allows external parties to verify that the user has the ability to spend the input.
+This specification defines the format for a proof of ownership which can be passed to a hierarchical deterministic wallet together with each input of an unsigned transaction. This proof allows the wallet to determine whether it is able to spend the given input or not. It also allows third parties to verify that a user has the ability to spend the input.
 
 ## Motivation
 
-In certain applications like CoinJoin and Lightning the hardware wallet has to deal with transactions containing external inputs. To calculate the actual amount the user is spending the hardware wallet needs to reliably determine whether the input belongs to the user or not. Without such a mechanism an adversary can fool the hardware wallet into displaying incorrect information about the amount being spent, which can result in theft of user funds.
+In certain applications like CoinJoin and Lightning, a wallet has to sign transactions containing external inputs. To calculate the actual amount the user is spending, the wallet needs to reliably determine for each input whether it belongs to the wallet or not. Without such a mechanism an attacker can fool the wallet into displaying incorrect information about the amount being spent, which can result in theft of user funds.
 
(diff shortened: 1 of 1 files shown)


andrewkozlik:
I should mention here that I also contemplated commiting to a particular UTXO and amount in the proof, but I wasn't able to find a strong reason to do so, see the design doc mentioned above. If we decide to do this in the future, then such a commitment should be optional because it would be required only in certain applications. It can be easily added using the flags: https://github.com/satoshilabs/slips/commit/b36692c2b67b535b8390acbf090de3ff2f9a9807 (+ updates needed to "Proof construction" and "Proof usage" sections).

andrewkozlik:
@instagibbs you might be interested in having a look at this too.

instagibbs:
I'll try to give it a read soon

slip-0019.md
```diff
@@ -5,24 +5,182 @@ Number:  SLIP-0019
...
+
+In case of *m*-of-*n* multi-signature scriptPubKeys the proof of ownership SHOULD contain the ownership identifiers of all *n* co-owners of that scriptPubKey. See [Identifier inclusion](#identifier-inclusion) for further details.
+
+A wallet MUST NOT produce and reveal an ownership identifier for a scriptPubKey which it does not control. Such a fake ownership identifier can be used to mount a denial-of-service attack.
+
+### Proof body
+
+The *proofBody* is a concatenation of the following fields:
+
+* *version* (4 bytes): b"\x53\x4c\x00\x19".
```
instagibbs, line 59: why these bytes?

slip-0019.md
```diff
@@ -5,24 +5,182 @@ Number:  SLIP-0019
...
+* *id*<sub>1</sub> || *id*<sub>2</sub> || ... || *id*<sub>*n*</sub> (32 bytes each): concatenation of the ownership identifiers for the given scriptPubKey, one for each co-owner, see [Identifier inclusion](#identifier-inclusion) for further details.
+
+### Proof footer
+
+The *proofFooter* is a concatenation of the following fields:
+
+* *scriptPubKey* (length-prefixed string).
+* *commitmentData* (length-prefixed string), any additional data to which the proof should commit, see below.
+
+The proof footer is not a part of the proof of ownership. It is included only in the *sighash* computation. Variable-length fields are encoded the same way as in Bitcoin transactions, as a length-prefixed string, where the length is encoded as a variable-length integer (VarInt).
```
instagibbs, line 75: > It is included only in the *sighash* computation.

Why? A brief justification should be in the text.

slip-0019.md
```diff
@@ -5,24 +5,182 @@ Number:  SLIP-0019
...
+
+## Proof usage
+
+### Verifying non-ownership of transaction inputs
+
+When a wallet is requested to sign a transaction, each external input SHOULD be accompanied with a proof of ownership so that the wallet may ascertain non-ownership of such an input in order to correctly inform the user about the amount they are spending in the transaction. For each external input the wallet takes the following steps:
+
+1. By reliable means obtain the scriptPubKey of the UTXO being spent by that input. Prior to SegWit version 1 witness programs this step involves acquiring the full transaction being spent and verifying its hash against that which is given in the outpoint.
+2. Parse the *proofBody*. If the *version* is not recognized or if any of the bits 1 through 7 of *flags* is set, then abort.
+3. Verify that the *proofSignature* is valid using the obtained scriptPubKey.
```
instagibbs, line 131: probably should re-mention that validation is done via bip322

slip-0019.md
andrewkozlik: 2f84a7a: Renamed it to versionMagic and added an explanation.

slip-0019.md
andrewkozlik: Done in 2f84a7a.

slip-0019.md
andrewkozlik: Done in 2f84a7a.

instagibbs:
appreciate the changes :+1:

andrewkozlik:
> appreciate the changes +1

Thanks for your input @instagibbs!

stepansnigirev:
@andrewkozlik, reads great! ACK :)

andrewkozlik:
> 1. Regarding PSBT, I think it's better to use `PSBT_GLOBAL_PROPRIETARY` and `PSBT_OUT_PROPRIETARY` with subtypes `PSBT_GLOBAL_PSBT_ID` and `PSBT_IN_PROOF_OF_OWNERSHIP` instead of introducing new types to the BIP 174 itself.

I wasn't sure about this myself, but in the end what prompted me to propose it as non-proprietary is the following statement in [BIP-0174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#proprietary-use-type):

> The proprietary use type is not to be used by any public specification [...]

Presumably SLIP-0019 can be considered a public specification. If the pull request to include this in BIP-0174 is rejected, then we can either decide to make it proprietary or, more likely, leave it out of this spec completely (which I am fine with doing even right now).

PR #940, opened 2020-05-29
SLIP-0173: Add Omni Human Readable Parts

msgilligan:
Omni is working on Segwit support and intends to require unique Bech32 addresses when sending  Omni Layer tokens on the Bitcoin chain. We would like to reserve the `o`, `to`, and `ort` Human Readable Parts.

The Human-readable parts are specified in OLE-300: https://github.com/OmniLayer/Documentation/blob/master/OLEs/ole-300.adoc#human-readable-part


dexX7:
I support this submission.

msgilligan:
Thank you!

PR #948, opened 2020-06-15
slip-0173: Add Oasis Network

tjanez:


slip-0173.md
```diff
@@ -48,6 +48,7 @@ These are the registered human-readable parts for usage in Bech32 encoding of wi
 | [Monacoin](https://monacoin.org/)              | `mona`     | `tmona` | `rmona`     |
 | [Myriad](https://myriadcoin.org/)              | `my`       | `tm`    |             |
 | [Namecoin](https://www.namecoin.org/)          | `nc`       | `tn`    | `ncrt`      |
+| [Oasis Network](https://oasisprotocol.org/)    | `oasis`    |         |             |
```
prusnak, line 51: No tesnet?

tjanez: We have a testnet, the upcoming one is called the [Oasis Amber Network](https://medium.com/oasis-protocol-project/introducing-the-amber-network-a-release-candidate-for-mainnet-7fe9c8d6a5c5), but it uses the same HRP as the Mainnet will.

I'll update the second column to indicate that both networks use the same HRP.

prusnak: It is unacceptable that both testnet and mainnet use the same prefix. Please fix this.

PR #950, opened 2020-06-25
add JPYS

dgyoshi:
Adds JPYS to slip-0044

SETTLENET JPY is a Japanese yen-denominated token that is used solely as method of payment for trading between the domestic and foreign entities that have underwent KYC/AML checks by Crypto Garage. SETTLENET JPY is issued as an Issued Asset on the Liquid Network, and is issued directly to the user’s Liquid Wallet. The underlying yen assets are fully segregated to a trust bank.

https://settlenet.io/
https://blockstream.info/liquid/asset/3438ecb49fc45c08e687de4749ed628c511e326460ea4336794e1cf02741329e

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -1052,6 +1052,7 @@ index | hexa       | symbol | coin
 1688  | 0x80000698 | BCX    | [BitcoinX](https://bcx.org)
 1729  | 0x800006c1 | XTZ    | [Tezos](https://tezos.com)
 1776  | 0x800006f0 | LBTC   | [Liquid BTC](https://blockstream.com/liquid/)
+1784  | 0x800006F8 | JPYS   | [JPY Stablecoin](https://settlenet.io/)
 1815  | 0x80000717 | ADA    | [Cardano](https://www.cardanohub.org/en/home/)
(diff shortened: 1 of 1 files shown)


prusnak:
Can you show me the source code of JPY wallet where you use the requested number as BIP44 path?

dgyoshi:
@prusnak Our wallet is not open source yet. JPYS asset has been issued on Liquid Network, so any wallets that are able to handle L-BTC will be used though.

prusnak:
Does it make sense to register SLIP-0044 for a Liquid Asset? Can these assets be stored on different keys than the underlying "gas" - Liquid BTC in this case? For Ethereum you need to store tokens on the same keys as the Ethereum, which is used for gas (paying the TX).

dgyoshi:
Yes, it does. It would be useful for assets created for fun because they might not need to register coin type. But it would be nice for JPYS to have specific coin type so that wallets will be able to discover assets without any difficulty.

prusnak:
You did not answer my question:

> Can these assets be stored on different keys than the underlying "gas" - Liquid BTC in this case?

dgyoshi:
Yes, assets on Liquid Network can be stored different keys to the one which L-BTC uses. And yes, L-BTC is used as fee in Liquid Network just like “gas” in the Ethereum.

PR #951, opened 2020-06-25
Slip-0044 add AUDL(#1618)

dpickering006:
Adds AUDL to slip-0044

AUDL is an Australian Dollar token issued on the Liquid network. It is used as an interexchange/OTC desk transport mechanism.

The underlying Australian Dollars are held in a local AUD bank account. 

Issuance and Redemption will occur at https://aud.one

The Liquid Asset ID is 

https://blockstream.info/liquid/asset/f59c5f3e8141f322276daa63ed5f307085808aea6d4ef9ba61e28154533fdec7


prusnak:
Does it make sense to register SLIP-0044 for a Liquid Asset? Can these assets be stored on different keys than the underlying "gas" - Liquid BTC in this case? For Ethereum you need to store tokens on the same keys as the Ethereum, which is used for gas (paying the TX).

PR #957, opened 2020-07-06
add HST

hschain:
slip-0044: add HST


prusnak:
Careful, HST was assigned number 532, not 531.

PR #965, opened 2020-07-31
slip-0044 : add wethio blockchain

zynecoin:
Homepage : [Wethio](https://wethio.io)

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -1144,6 +1144,7 @@ index | hexa       | symbol | coin
 49344 | 0x8000c0c0 | STASH  | [STASH](https://stashpay.io/)
 52752 | 0x8000ce10 | CELO   | [Celo](https://celo.org/)
 65536 | 0x80010000 | KETH   | [Krypton World](http:/krypton.world/)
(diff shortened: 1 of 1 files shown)


prusnak:

zynecoin:
@prusnak we checked the coins list on the repository and decided to go with the coin type 77777' as it was not being used.

Currently our blockchain and wallet app are running on mainnet , and people are holding ZYN currency in their wallets, so we request you to please allow us using this coin type as it is not registered currently also.

prusnak:
Show me the code of the wallet where 77777' is used

zynecoin:
@prusnak  Currently our wallet respository is private as work on some of the modules is going on, but you can check our wallet website (https://wallet.wethio.io) where we specify the hdPath in the ledger login option, you can check from our documentation too at https://docs.wethio.io/developers-guide/wethio-network 

Also we have created a PR for the ledger integration of our blockchain, which further details the same path - https://github.com/LedgerHQ/app-ethereum/pull/104

If this does not work for you, we can add you as a contributor to check the code from our repo or else will try to make the code public in some time

PR #983, opened 2020-09-02
Add XDS coin

dangershony:


prusnak:

dangershony:
Ah unfortunately this number is already in use by the miners, network was launched a few month back.

prusnak:
Not sure how the BIP44 path relates to miners.

dangershony:
Miners use wallets for POS and derive addresses to hold coins (I should have said stakers)
Sure we can change our wallets but it will represent bigger effort so I'd just abandone this PR instead (I was not aware we should stick to added numbers).
No harm done if you won't merge like this.

PR #995, opened 2020-10-06
SLIP-44 Added IRISnet

iRhonin:
Added IRISnet token.


prusnak:

iRhonin:
@prusnak done.

PR #1001, opened 2020-10-08
slip-0039: Explain design rationale regarding compatibility with BIP-39

andrewkozlik:
I noticed that a lot of people are surprised when they find that SLIP-39 shares are not convertible to a BIP-39 mnemonic, so I decided to emphasize that in the abstract and explain it in the design rationale. Also, some individuals have expressed a concern that the lack of this feature will lead to vendor lock-in. I tried to explain why that concern is unwarranted, so that we may easily reference this in blog posts or discussions. I tried to keep it succinct. Just have a quick look to confirm our stance.


Luckylukee:
`BIP39` mnemonic, or generate a new random one. Typing your own twelve words will probably not work how you expect, since the words require a particular structure (the last word contains a checksum).
More info
BIP39 Mnemonic code for generating deterministic keys
Read more at the official #1001 BIP39 spec

PR #1007, opened 2020-10-14
add haidai.network KELP

jiguangin:
add haidai.network KELP


prusnak:

PR #1016, opened 2020-11-11
Re-use 15 

gjhiggins:
Reassign from extinct ClearingHouse to Gapcoin


prusnak:
I see no reason why to do this. Allocate the new number.

PR #1020, opened 2020-11-17
Add BCHA

alisawoodenate:
A BCHA path code should be there for wallet developers.


prusnak:
This does not make sense, since neither BCHA nor BCHN implemented strong replay protection and thus they are still the same chain.

PR #1049, opened 2021-01-10
slip-0044: rebrand STB

shanev:
STB rebranded as STARS. STB was never deployed to mainnet.


prusnak:
Please show me a proof that STB was rebranded to STARS.

It seems the old web still runs at the old domain. I would expect there would be a redirect from the old domain to the new domain, or some kind of announcement of the rebrand.

shanev:
Sorry, submitted the PR too early. 

The redirect should work now (stakebird.com -> stargaze.fi). 

Announcement of the rebrand is at : https://twitter.com/starsdefi/status/1347793462951895041?s=20, and https://publicawesome.com.

PR #1059, opened 2021-02-06
Rebranding of SnowGem to TENT

ciripel:
Lately SnowGem rebranded to TENT so this is just an actualization of the coin name.


prusnak:
Merged as https://github.com/satoshilabs/slips/commit/65ed453f8de5a5dc77b6bc850942f636fef67e13

PR #1061, opened 2021-02-09
Remove Lightning

MaxHillebrand:
In 'regular' Lightning Network flow [channel openings / routed payments / cooperative / forceful channel closes], there are no multi-user multi-input transactions, afaik.

There are however channel openings where both peers provide inputs to the opening transaction [I'm blanking on the correct term here...] as well as channel factories among more than 2 users. These type of transactions are CoinJoins to open Lightning channels, but not 'Lightning itself'.

I think, that thus we can say that SLIP 19 is only relevant for CoinJoin, and not for Lightning.


andrewkozlik:
My apologies for not getting around to looking at this PR earlier. I seem to have missed it. I think the term you were looking for is a "funding transaction for a dual-funded channel". I wouldn't call that a CoinJoin. In my mind a CoinJoin transaction is a joint transaction where all parties receive the same amount out of the transaction as they put in (minus fees) and the purpose of the transaction is to mix the UTXOs of all participants to obfuscate UTXO ownership to external observers. However I agree that otherwise in regular Lightning Network flow there is no need for external inputs. I created a new PR https://github.com/satoshilabs/slips/pull/1286 where I am changing the wording to
> In certain applications like CoinJoin or opening a dual-funded channel in the Lightning Network, a wallet [...]

I am closing this PR, since it's targeting a stale branch. We can continue the discussion in https://github.com/satoshilabs/slips/pull/1286.

MaxHillebrand:
Good that you clean up and close, thanks.

Just a nit, but the "official" (at least gmax 2013) definition of a coinjoin is multi user input consolidation, and not equal output amounts. Look at PayJoin, which is commonly referred to as a CoinJoin.

PR #1062, opened 2021-02-09
Remove that CJ must be P2WPKH

MaxHillebrand:
> even though CoinJoin transactions currently work only with P2WPKH addresses

That is not correct, there are multiple CoinJoin implementations that support P2PKH / P2SH address types. For example old implementations like Dark Wallet / Blockchain .com, and even current implementations like Joinmarket.

Only Wasabi & Samourai are P2WPKH only.

Fwiw, segwit fixes the transaction malleability, and thus enables to register unconfirmed coins in subsequent CoinJoins. [we've recently deactivated registering unconfirmed CJ outputs at the zkSNACKs Wasabi coordinator]


MaxHillebrand:
I have a small question not worth opening a new issue... 

> In case of CoinJoin transactions the commitmentData SHOULD contain a globally unique PSBT identifier (psbtId).

could this be the unique RoundID? though the RoundID is generated BEFORE the transaction is built by the coordinator, thus it seems different than the PSBTID.

xianggithub01:
你们说的都有道理。

xianggithub01:
You all have a point. Please speed up my account, thank you 19115936551

andrewkozlik:
I agree with the change, but this PR is targeting a stale branch and changing it to `master` messes up the diff. I made the same change in https://github.com/satoshilabs/slips/pull/1286, which is why I am closing this PR. Thanks for your input @MaxHillebrand.

> I have a small question not worth opening a new issue...
> 
> > In case of CoinJoin transactions the commitmentData SHOULD contain a globally unique PSBT identifier (psbtId).
> 
> could this be the unique RoundID?

Yes, exactly. In particular I think in Wasabi 2 it was decided that it would be the coordinator name encoded as a [length-prefixed string](https://en.bitcoin.it/wiki/Protocol_documentation#Variable_length_string) followed by 32 bytes of the round ID. 

> though the RoundID is generated BEFORE the transaction is built by the coordinator, thus it seems different than the PSBTID.

I don't see a problem. The psbtID can be any nonce acting as a unique label, so the round ID is perfectly fine. The value does not have to be derived from the PSBT or the transaction data. See https://github.com/satoshilabs/slips/blob/master/slip-0019.md#additional-commitment-data.

PR #1063, opened 2021-02-09
Replace AskCoin by DFINITY

Dfinity-Bjoern:
As discussed with Artia: AskCoin has been abandoned, while DFINITY has use of this path in the wild.


blocklee:
It's shameless to rob other people's IP. It looks like DFINITY is very low. Although the AskCoin project has been suspended, it does not mean that others will not use it.

PR #1082, opened 2021-03-20
add/register Raptoreum bip44 index 10226

npq7721:
add/register Raptoreum bip44 index 10226


prusnak:
use lowercase for hex digits

npq7721:
fixed

PR #1095, opened 2021-04-09
Update slip-0044.md

EVRICE:


EVRICE (commented):
Evrice web address fix

PR #1098, opened 2021-04-12
add TxRef's HRPs from BIP-136 to SLIP-173

danpape:
This adds HRP's that are used in BIP-136 to SLIP-173. I have created a new section as our use does not relate to witness programs. Please let me know if you feel this isn't a good way of organizing things.


prusnak:
I don't think this really belongs here. It's not the goal of SLIP-173 to map every possible usage of Bech32 in the wild.

PR #1105, opened 2021-04-23
Register DECENOMY coins types for BIP-0044

decenomy:
We ask to register unique codes for our coins on the slip-0044.md to increase the correctness of our code and future use of Trezor on all our coins


decenomy:
Thank you :)

PR #1114, opened 2021-04-30
add Meter

philipappiah:
add Meter coin type at 18000


prusnak:
18000 in hex is 0x4650 not 0x4560

PR #1121, opened 2021-05-13
Register DECENOMY coin MONK on SLIP-44 for BIP-0044

decenomy:
We ask to register a unique code for our coin MONK on the slip-0044.md for future use on Trezor wallets

Thanks in advance


pedro-at-decenomy:
Thank you @prusnak :)

PR #1148, opened 2021-08-18
Add Cosmos Ecosystem chains to SLIP 173

sunnya97:
Hi, I'm one of the core devs of the Cosmos ecosystem.  We actually use bech32 addresses as the default for all the chains in our ecosystem.  I just realized that only very few of the chains in Cosmos have registered themselves in the SLIP-173 repo, so I've gathered the HRP info of as many of them I could find, and added them.

The one deleted line is for "IOV", which is a stale entry because they have rebranded to Starname (which I included).  See: https://medium.com/iov-internet-of-values/iov-has-become-starname-a95e27165928

### DIFF
--- a/slip-0173.md
+++ b/slip-0173.md
@@ -21,86 +21,85 @@ The BIP repository does not want to deal with assigning the values for various c
 
 These are the registered human-readable parts for usage in Bech32 encoding of witness programs.
 
-| Coin                                           | Mainnet       | Testnet | Regtest     |
-| ---------------------------------------------- | ------------- | ------- | ----------- |
-| [Agoric](https://agoric.com/)                  | `agoric`      |         |             |
-| [Akash Network](https://akash.network/)        | `akash`       |         |             |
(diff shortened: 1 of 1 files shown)


sunnya97:
Btw, there's on Cosmos chain (Likecoin) that apparently did not change their HRP prefix from the Cosmos SDK default of `cosmos`, and thus its addition would break the current 1:1 mapping of chain to prefix.  Is this desirable?

prusnak:
> Is this desirable?

No. We will not add colliding entries to the list.

sunnya97:
@prusnak Oh in that case we need to remove Likecoin (which was included in this PR)

prusnak:
> @prusnak Oh in that case we need to remove Likecoin (which was included in this PR)

Fixed in dac71b55259bb3619eac95cd3255a830215529bf

sunnya97:
@prusnak done!  Line 76

PR #1172, opened 2021-10-18
Update slip-0044.md

samisbakedham:
Replaced LAND, labeled as "dead", with Nobility, an algroand-based cryptocurrency


prusnak:
slip44 numbers are not meant to be reused

please use a different number

PR #1202, opened 2021-11-25
Added Avalanche C-Chain

briansoule:
Added Avalanche C-Chain EVM compatible chain, which is distinct from the Avalanche P & X-Chain predecessors.


prusnak:
please use the link pointing to the coin, not to a block explorer

PR #1203, opened 2021-11-26
Added Binance Smart Chain

briansoule:


prusnak:
What is the difference between this and already registered BNB?

briansoule:
Binance Chain I believe is a Tendermint based chain, while Binance Smart Chain is an EVM compatible chain. Which is understandably confusing because they both use BNB.

https://support.forsage.io/hc/en-us/articles/360020637179-Difference-between-Binance-Chain-BNB-BEP-2-format-and-Binance-Smart-Chain-BNB-BEP-20-format-

prusnak:
please use BSC abbreviation (which seems is the correct one)

briansoule:
That was my first inclination, however there seems to be a dead chain called Bitcoin Smart Contract that also uses BSC at position 519.

prusnak:
I know, but that does not mean you should come up with your own abbreviation.

sunnya97:
Why is BSC being added a SLIP-044 entry?  As far as I'm aware, it's an EVM chain that uses address compatibility with Ethereum, and thus reuses the Ethereum path of `60`, and thus should not be assigned a new SLIP-044 path.  Ditto for the chains added in #1201

PR #1204, opened 2021-11-26
Add ARCHEthic's UCO

drvarun:
Add ARCHEthic Blockchain's Universal Coin UCO
https://archethic.net/#scroll-to-uco-section


prusnak:
Isn't UCO just an ERC-20 token on Ethereum?

drvarun:
We have released our alpha-mainnet, with beta-mainnet around the corner. Beta-mainnet will also have the support for token-swap.

PR #1223, opened 2021-12-22
Fix signatures in SLIP-0019 test vectors

andrewkozlik:
Related to https://github.com/trezor/trezor-firmware/pull/2034.


onvej-sl:
What exactly do you expect from me in this review? :)

andrewkozlik:
> What exactly do you expect from me in this review? :)

Since you wrote an independent implementation of SLIP-0019 in Wasabi Wallet, which you will be fixing now, I think it would make sense to check that the test vectors are correct by running them against your implementation (or at least those, that you can).

onvej-sl (approved):
I checked that Wallet Wasabi
  * generates the same ownership proof as in the P2WPKH test vector and
  * successfully verifies the ownership proof in the P2TR test vector.
 
This is all I can easily do since
  * Wallet Wasabi supports only P2WPKH and P2TR script types and
  * it generates different P2TR ownership proofs since it generates nonces in Schnorr signatures in a different way.

PR #1240, opened 2022-01-08
SLIP10: P-256-clarification

tarakby:
The document mentions that extending BIP32 algorithms from secp256k1 to NIST P-256 only requires substituting the curve order (n) of secp256k1 by the order of P-256. 

While this is correct in the case of Private parent key → hardened private child key, more curve domain parameters need to be substituted for :
 - Private parent key → non-hardened private child key  (scalar point multiplication is needed)
 - Public parent key → non-hardened public child key (scalar point multiplication and point addition are needed)
In these 2 cases, it is required to also substitute the prime field (F_p), the base point (G), and the Weierstrass equation parameter (a). 

The proposition updates the document to mention the domain parameter change and highlight the primitives that need be updated to the new curve. This should avoid possible confusion in the implementations.

### DIFF
--- a/slip-0010.md
+++ b/slip-0010.md
@@ -40,12 +40,15 @@ for different elliptic curves with different orders.
 We adapt the master key generation from BIP-0032.  To use different
 private keys for different curves we use different keys for the HMAC
 hash that generates the master key.  For the NIST P-256 curve the only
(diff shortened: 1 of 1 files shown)


Lucid369 (approved):

andrewkozlik (approved):
LGTM

PR #1244, opened 2022-01-15
add new base58 prefixes for Yilacoin

JayBitron:


prusnak:
What is the text representation of the prefixes? They are missing from the table.

JayBitron:
added

PR #1260, opened 2022-02-01
added cudos to slip-0173

V-Staykov:


prusnak:
You can't have the same prefix for mainnet and testnet

V-Staykov:
I've fixed this.

PR #1265, opened 2022-02-08
Add two coins from the SeniorBlockchain

miladsoft:


prusnak:
use lowercase letters for hex digits

miladsoft:
thank you . I corrected them.

PR #1266, opened 2022-02-08
slip-0044: add coin

JayBitron:


prusnak:
the domain does not work

JayBitron:
> the domain does not work

It works.

PR #1278, opened 2022-02-23
Fix velas derivation path

vldm:
We found that in this commit https://github.com/satoshilabs/slips/commit/1140ce26a621a23cd99460497e08aa4aab6e9f8c
Someone add VLX to the list.

In velas we use different constant https://github.com/velas/web3t/blob/master/providers/solana.js#L219
(Note: module solana because our chain is based on solana fork)


prusnak:
Please use the following if Velas is not using the assigned value:

```
574   | 0x8000023e | ---    | reserved
```

vldm:
@prusnak fixed, sorry for delay.

PR #1296, opened 2022-04-08
add slip-0044 cube infos

hcgaptx:
add cube infos


prusnak:
Remove Test token. 

Testnet tokens should use 1 as slip44 value.

PR #1300, opened 2022-04-13
Update slip-0173.md: Add Oraichain

ducphamle2:


prusnak:
1. keep the list sorted
2. you can't have the same prefix for testnet

PR #1326, opened 2022-05-21
slip-0044: remove coin

milapsheth:
Axelar decided to use the Cosmos ecosystem default of `118` for `AXL`. Reference in the [cosmos chain registry](https://github.com/cosmos/chain-registry/blob/master/axelar/chain.json#L17).


prusnak:
Closed in e4c73980086764da3336578879a68de8260bc670

PR #1328, opened 2022-05-24
Add NxtMeta

jsong1230:
NxtMeta EVM network wil use 901 as a BIP44 coin_type


prusnak:
please add URL

prusnak:
Already added

PR #1330, opened 2022-05-25
slip-0044: use 898 for Assetchain

linj-disanbo:
use 898 for Assetchain (AS)


prusnak:
please fix alignment - there is extra space before "AX"

linj-disanbo:
> please fix alignment - there is extra space before "AX"

fixed

PR #1331, opened 2022-05-26
Add Slashtags to SLIP-0044

Nazeh:
[Slashtags](https://github.com/synonymdev/slashtags) are keyPairs acting as decentralized identifiers and addresses for further metadata over data structures optimized for p2p sharing.

Slashtags' keyPairs can be derived from a `primaryKey` and a `name`. This `primaryKey` can and should be derived from a Bitcoin seed.

The derivation scheme is described [here](https://github.com/synonymdev/slashtags/blob/master/specs/slashtags-key-derivation.md#from-seed-to-primarykey).

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -729,7 +729,7 @@ Coin type | Path component (`coin_type'`) | Symbol | Coin
 698   | 0x800002ba | VEIL   | [Veil](https://www.veil-project.com)
 699   | 0x800002bb | GIO    | [Gio](https://github.com/qGio)
 700   | 0x800002bc | XDAI   | [xDai](https://blockscout.com/poa/dai)
-701   | 0x800002bd |        |
+701   | 0x800002bd | COM    | [Commercio.network](https://commercio.network)
 702   | 0x800002be |        |
 703   | 0x800002bf |        |
 704   | 0x800002c0 |        |


prusnak:
> can and should be derived from a Bitcoin seed.

Why should these be derived from `m/44'` and not some Synonym-specific path?

Nazeh:
Yes, I assume that Slashtags modules won't have direct access to `m` and would instead get a `primaryKey` from the hosting wallet [for more security](https://github.com/synonymdev/slashtags/tree/master/packages/sdk#security).

This PR hopes to help users generate that `primaryKey` consistently across wallets, and thus restore their Slashtags identities and attached metadata.

prusnak:
> Yes, I assume that Slashtags modules won't have direct access to `m` an

I thought so. But I your assumption has flaw. If you wanted proper domain separation, you'd give `m/44'/0'` or even `m/44'/0'/0'` to a signer, not `m/44'`. Since this is not happening and you are introducing a completely new purpose , you might as well use the new purpose path, e.g. `m/123456'`, because your key usage has nothing to do with regular BIP-44 accounts.

Nazeh:
@prusnak Thanks. But Is there a way to register a purpose path rather than coin types?

prusnak:
> But Is there a way to register a purpose path rather than coin types?

Quoting from [BIP43](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki):

> We encourage different schemes to apply for assigning a separate BIP number and use the same number for purpose field, so addresses won't be generated from overlapping BIP32 spaces.
>
> Purpose codes from 10001 to 19999 are reserved for [SLIPs](https://github.com/satoshilabs/slips).

So you can either try to register your purpose as a BIP or as a SLIP.

Nazeh:
Thanks for the clarification.

PR #1336, opened 2022-06-02
Update slip-0044.md

benben2019:
Update PlatON and Alaya info.


benben2019:
Update PlatON and Alaya info.

PR #1347, opened 2022-06-20
Chronic Chain Bech32 Prefix

discoverdefiteam:


JeremyParish69:
Not quite in alphabetical order, should go up a few more rows, just below chihuahua

discoverdefiteam (commented):
whoops! fixed 👍

PR #1350, opened 2022-06-24
Added Telos to Slip.44

nathanduft44:


prusnak:
Please use 995.

994 is already used by another coin - see https://github.com/satoshilabs/slips/pull/1349

nathanduft44:
@prusnak Oops thanks for pointing that out. Fixed

PR #1355, opened 2022-07-03
Remove Sentinel registration of 811

JeremyParish69:
Remove Sentinel's registration of CoinType 811. They use 118.


JeremyParish69:
May I ask why this hasn't been changed? Do we need someone from the team to confirm herein?

PR #1363, opened 2022-07-18
feat: Add FLR to slip-0044 as coin 30,001

aes256Slater:


RareData:
Do not use coin type 30001, all existing integrations for Flare Network uses coin type 60. See also: https://github.com/satoshilabs/slips/pull/1452

aes256Slater:
> Do not use coin type 30001, all existing integrations for Flare Network uses coin type 60. See also: #1452

60 conflicts with an existing chain, Ethereum, which could be quite problematic

RareData:
I represent the Flare team. Re-using coin type 60 is intentional and widely
done by other EVM-chains including Polygon, BNB Smart Chain and Songbird.

Both the Songbird (SGB) and Flare (FLR) token distributions are done to
addresses derived using coin type 60. All widely used Songbird and Flare
wallet implementations use coin type 60. We have no intention to move to a
new coin type.

PR #1366, opened 2022-07-21
Adds CHZ - Chiliz

RilRil:
Hello, I would like to register our CHZ token.
Thanks in advance


prusnak:
@RilRil The values you provided (decimal and hex value) do not match. Which one is the correct?

RilRil:
> @RilRil The values you provided (decimal and hex value) do not match. Which one is the correct?

indeed.. dont know why. you can remove the CHZ entry actually.
thank you

PR #1374, opened 2022-08-08
add Crescent

blockchainisnot:


prusnak:
testnet cannot have the same prefix as mainnet

blockchainisnot:
I'm sorry. I just removed it

PR #1384, opened 2022-08-16
Split Terra into Classic and 2.0

JeremyParish69:


prusnak:
There's no need to have duplicate entries.

JeremyParish69:
Thanks for considering. I wrote a script that uses the names here (pretty name on chain reg) as a lookup key, but if we don't want any dupes then I'll just write in an exception for Terra this once.

PR #1393, opened 2022-08-29
BIP-32 for BLS signature

b00f:
This PR standardizes a way to derive hardened and non-hardened keys for BLS signature based on BIP-32.

CKD functions are almost similar to the CDK functions defined in BIP32 with two changes:

1. The 0x00 padding for the private key is removed. The 0x00 pads the private key to make it 33 bytes long, which is the same length as the compressed public key for secp256k1. In BLS this is not the case, because the public key is either 48 or 96 bytes.
 
2. The parse<sub>256</sub>(I<sub>L</sub>) ≥ n condition is removed, otherwise more than 50 percent of the keys will be invalid.

Other parts of the functions are same as BIP-32 CDK functions.


prusnak:
Why not expanding the already existing SLIP-0010?

b00f:
[@prusnak](https://github.com/prusnak)

SLP-0010 is a really good option for most of the elliptic curves, but for[ pairing friendly curves](https://datatracker.ietf.org/doc/html/draft-yonezawa-pairing-friendly-curves-02) probably is not a good option. For example for BLS12-381 curve:

1. Step 5 in both CKDpriv and CKDpub functions will cause more than 50 percent of keys to become invalid. Note that curve order is: 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001
2. Two public keys can be calculated per private key. It depends on which curve the public key is defined.So step 1 needs to be extended more.
3. Public key in not 33 bytes. It is 48 or 96 bytes. So 0x00 padding doesn't work here.

Sahismail:
0xf4d856ce523faa78ff2675966428a98790b3ad1c

b00f:
In the last commit, test vectors are updated. In order to be more general, the master key in HMAC function changed from "Zarb seed" to "BLS12381-HD-KEYCHAIN":

https://github.com/pactus-project/pactus/pull/386

b00f:
@prusnak Did you check the spec? If there is any questions, please let me know.

LisaBurke40:
Yea

b00f:
> Yea

@LisaBurke40 Any update about this PR?

bitcoinfinancier (approved):

prusnak:
There is no immediate plan to include BLS in SatoshiLabs products and therefore we don’t have enough resources to review this document and we do not need to have this document in this repository. 

I suggest you find different repository where you can try to standardize your effort. 

AFAIK [Fedimint](https://github.com/fedimint) uses BLS, so you might want to check your work with them.

PR #1399, opened 2022-09-06
add ETHW to slip-0044

sup3ran0n:


prusnak:
Link to the source code where this value is used?

butogon (changes requested):
Thanks

PR #1417, opened 2022-10-02
add stargaze testnet

yubrew:


prusnak:
Testnet shouldn't have the same prefix as mainnet.

PR #1425, opened 2022-10-19
Adding Peer coin in 718 (previously unreserved)

lawmeskiviahs:
718 for Peer


prusnak:
Peercoin is already at number 6

PR #1428, opened 2022-10-23
Update slip-0044.md

develoQ:
update XRP coin name to XRP


prusnak:
I think it's better to keep it as is, so people are able to find the coin using the old name.

develoQ:
"Ripple" is a specific company name, not the name of the native token of the XRP Ledger, which is a public distributed ledger.

It would be undesirable for the wrong name to remain in use in the future as information for the standard.

PR #1432, opened 2022-10-27
Adding jackal

TheMarstonConnell:


prusnak:
1. keep the list sorted
2. testnet cannot have the same prefix as mainnet

TheMarstonConnell:
> 1. keep the list sorted
> 2. testnet cannot have the same prefix as mainnet

Should be fixed now then. Sorry about that

PR #1448, opened 2022-11-17
Add KILT live and testnet to slip-0044 coin type list.

trusch:
This will add the KILT networks to the slip-0044 coin type list. KILT Spiritnet (which is the live network, a parachain of polkadot) has the same coin type number as it's parachain ID, following the example of Exosama Network and others. Our testnet has that number + 1.


prusnak:
> ur testnet has that number + 1.

No, all testnets should always use the number 1. Please change the PR accordingly.

PR #1456, opened 2022-11-29
Add StaFiHub to SLIP-0173

Tore19:


prusnak:
Mainnet and testnet cannot use the same prefixes.

Tore19:
> Mainnet and testnet cannot use the same prefix.

Got it. Just removed the testnet prefix. Thanks for your feedback!

PR #1481, opened 2023-01-16
Add Nexa coin

vgrunner4v:


prusnak:
Use the lowercase hex letters

There is no such thing as P2PKT

vgrunner4v:
@prusnak - hex fixed. about P2PKT, please see: https://spec.nexa.org/doc/address

PR #1482, opened 2023-01-17
Changing 'Peer Meta Coin' name to 'Peer Coin'

lawmeskiviahs:


prusnak:
Peercoin (PPC) is already assigned SLIP-0044 number 6.

I am not merging this so we won't confuse users which coin is which.

urkishan:
Hi @prusnak, As the currency code for the coin is `PEER` and the name is `Peer Coin` and it is totally different blockchain.

urkishan:
So if there is anything else we can do please let us know.

prusnak:
You can rename your coin so that it doesn't feed on the name of another coin that has been around since 2012.

PR #1484, opened 2023-01-26
added bech32 prefix for quicksilver chain

ajansari95:


prusnak:
don't touch the line above

PR #1494, opened 2023-02-12
slip-0019: invalid proof body for the second test vector

louneskmt:
The magic version bytes aren't complete, and as the user confirmation is set to `True`, the flag byte should be `01`. The splitted components are okay, these are just two typos in the full proof.


prusnak (approved):
It seems you are right. I will let Andrew to do another pair of eyes review.

louneskmt:
I'm also not able to get the correct `sighash` while I can with the same code for the first test vector. Not sure if that's on me or not (maybe a wrong handling of the `commitment_data`, that was just empty in the first test). I would greatly appreciate if you could quickly double check that too!

andrewkozlik:
Thanks for letting us know @louneskmt. The sighash was wrong too, precisely because of the commitment data and user confirmation flag, which then also results in a different signature. So I created a separate PR fixing all of those issues. Can you please have a look at https://github.com/satoshilabs/slips/pull/1495 and confirm the corrected values? I checked the remaining test vectors and they seem OK.

louneskmt:
Hey @andrewkozlik! Thanks a lot, I now get the same `sighash`. I can't confirm for the signature as I haven't implemented that part yet tho.

andrewkozlik:
So I propose that we close this PR in favor of #1495. Once you have implemented the signature checking, please write a comment in #1495 whether you find the test vector correct or not. Thanks!

louneskmt:
All good, closing in favor of #1495.

PR #1495, opened 2023-02-13
slip-0019: Fix test vector 2.

andrewkozlik:
Replaces https://github.com/satoshilabs/slips/pull/1494

[Test vector](https://github.com/satoshilabs/slips/blob/andrewkozlik/slip19b/slip-0019.md#test-vector-2-p2wpkh-nested-in-bip16-p2sh)


louneskmt:
Shouldn't the `scriptSig` part of the proof be `160014e0cffbee1925a411844f44c3b8d81365ab51d036` instead of `17160014e0cffbee1925a411844f44c3b8d81365ab51d036`? It looks to me that you double prefixed it with lenght.

The 20-byte `keyhash` if `e0cffbee1925a411844f44c3b8d81365ab51d036`, and the P2SH `redeemScript` for a P2SH-P2WPKH is `0x0014{20-byte keyhash}`. So we have `redeemScript=0014e0cffbee1925a411844f44c3b8d81365ab51d036`. For spending of P2SH-P2WPKH, we must have `scriptSig=redeemScript`, so `scriptSig=0014e0cffbee1925a411844f44c3b8d81365ab51d036`. 

Finally, prefixing `scriptSig` with its VarInt lenght, we should have `160014e0cffbee1925a411844f44c3b8d81365ab51d036`.

Please correct me if I'm wrong.

andrewkozlik:
> Shouldn't the `scriptSig` part of the proof be `160014e0cffbee1925a411844f44c3b8d81365ab51d036` instead of `17160014e0cffbee1925a411844f44c3b8d81365ab51d036`? It looks to me that you double prefixed it with length.

Yes, you have a valid point, the scriptSig should be like you say `1600...`. But to be clear, I think the proof of ownership is correct the way it is. The byte 0x17 is the varint length of the scriptSig and the first byte of the scriptSig 0x16 is the [opcode](https://en.bitcoin.it/wiki/Script#Constants), which pushes the next 22 bytes onto the stack. You can see the same thing in [this transaction](https://btc1.trezor.io/tx/f3903da832a624463b39aa19e07580a765e5d6081ea89be7825b9dafdff5ce4d) that I randomly picked off the blockchain:
image

> The 20-byte `keyhash` is `e0cffbee1925a411844f44c3b8d81365ab51d036`, and the P2SH `redeemScript` for a P2SH-P2WPKH is `0x0014{20-byte keyhash}`. So we have `redeemScript=0014e0cffbee1925a411844f44c3b8d81365ab51d036`.

Yes, this is correct.

> For spending of P2SH-P2WPKH, we must have `scriptSig=redeemScript`, so `scriptSig=0014e0cffbee1925a411844f44c3b8d81365ab51d036`.

Here I disagree. See [BIP 141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#witness-program), Case 2 in the Witness program section: "The scriptSig must be exactly a push of the BIP16 redeemScript". So
```
redeemScript=0014e0cffbee1925a411844f44c3b8d81365ab51d036
scriptSig=160014e0cffbee1925a411844f44c3b8d81365ab51d036
lengthPrefixedScriptSig=17160014e0cffbee1925a411844f44c3b8d81365ab51d036
```

If you look at the scriptSig of Test vector 3, you will see a similar thing:
```
6b // VarInt length of scriptSig (107 bytes)
48 // pushes the next 72 bytes onto the stack
3045022100e818002d0a85438a7f2140503a6aa0a6af6002fa956d0101fd3db24e776e546f0220430fd59dc1498bc96ab6e71a4829b60224828cf1fc35edc98e0973db203ca3f001
21 // pushes the next 33 bytes onto the stack
02f63159e21fbcb54221ec993def967ad2183a9c243c8bff6e7d60f4d5ed3b3865
```

Again, you are correct that `6b` is not part of the actual scriptSig, but it is the length prefix and has to be in the proof of ownership, otherwise we wouldn't know where the scriptSIg ends, because in general there can be multiple push operations like in the case of P2PKH. What's confusing about Test vector 2 or in general about P2WPKH nested in BIP16 P2SH is that there is only one push of data onto the stack, which makes it look double-prefixed.

andrewkozlik:
I added a [commit](https://github.com/satoshilabs/slips/pull/1495/commits/82dfda7ea2c5f8c2de02cb29fbacaa6584934244) which splits the scriptSig length from the actual scriptSig.

louneskmt:
Oh that's why, I forgot the opcode to push the data. Thanks a lot for the detailed explanation, it makes a lot more sense now. I'm able to pass the test vectors, all good! You may merge this 😁

andrewkozlik:
Thanks @louneskmt!

PR #1504, opened 2023-02-23
Add Age Yubikey identities

lourkeur:
The [Yubikey plugin for Age](https://github.com/str4d/age-plugin-yubikey) uses NIST P-256 (or P-384) keypairs. This is different than the native Age keypairs which are X25519. Having a SLIP-0044 would allow derivation and also ENS integration, following the model of https://github.com/ensdomains/address-encoder/pull/378.


prusnak:
Can you rename `Age Encryption` to `Age X25519 Encryption` and `Age Yubikey Encryption` to `Age NIST Encryption`?

PR #1505, opened 2023-03-01
Update slip-0044.md

coranos:
Someone finally checked and it turns out Banano was spelled wrong (should not have an S at the end), and the symbol is BAN not BANANO/


prusnak:
Please fix your commit, you broke the formatting of your change completely.

coranos:
Fixed. That was strange.

PR #1506, opened 2023-03-05
Add mewc

michaelmcshinsky:
[meowcoin](https://www.mewccrypto.com/)

[https://github.com/JustAResearcher/Meowcoin/blob/master/src/chainparams.cpp](https://github.com/JustAResearcher/Meowcoin/blob/master/src/chainparams.cpp)


prusnak:
keep the list sorted by number

michaelmcshinsky:
> keep the list sorted by number

Thanks, fixed :)

PR #1521, opened 2023-03-23
Register Coreum address prefixes

wojtek-coreum:


prusnak:
Don't reformat the whole table, just add one new line

wojtek-coreum:
@prusnak 

Done

PR #1523, opened 2023-03-25
added noble

danbryan:
Noble mainnet launches Monday, we need this in place to get on the chain registry.

https://github.com/cosmos/chain-registry/pull/1705

https://blockworks.co/news/noble-solve-cosmos-liquidity


prusnak:
Please don't use the same prefix for mainnet and testnet.

PR #1533, opened 2023-04-19
Update slip-0044.md

cqcwasd:
add WaltonChain on slip-0044


cqcwasd:
add WaltonChain on slip-0044

prusnak:
The values you provided (decimal and hex value) do not match. Which one is the correct?

prusnak:
@cqcwasd  The values you provided (decimal and hex value) do not match. Which one is the correct?

PR #1535, opened 2023-04-23
added BitBay slip44 path

anoxxxy:
https://bitbay.market

https://github.com/bitbaymarket/bitbay-core

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -1124,7 +1124,7 @@ All these constants are used as hardened derivation.
 | 2086       | 0x80000826                    | KILT    | KILT Spiritnet                    |
 | 2109       | 0x8000083d                    | SAMA    | Exosama Network                   |
 | 2112       | 0x80000840                    | IoTE    | IoTE                              |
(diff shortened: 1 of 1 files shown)


slip-0044.md
```diff
@@ -1124,6 +1124,7 @@ All these constants are used as hardened derivation.
 | 2086       | 0x80000826                    | KILT    | KILT Spiritnet                    |
 | 2109       | 0x8000083d                    | SAMA    | Exosama Network                   |
 | 2112       | 0x80000840                    | IoTE    | IoTE                              |
+| 2125       | 0x80000853                    | BAY     | BitBay                            |
```
prusnak, line 1127: 0x853 is 2131
2125 is 0x84d

so which one is it?

prusnak (changes requested):

slip-0044.md
anoxxxy: sorry, it should be  `2125`, so `0x84d` is correct.

PR #1538, opened 2023-04-26
Adding Ekta chain entry on slip-44

himang305:


himang305:
Adding Ekta chain entry on Slip_0044 at 656 Coin Type

PR #1539, opened 2023-04-27
Register BIP-0044 coin type for celestia.

HoytRen:
https://celestia.org
https://github.com/celestiaorg


prusnak:
all testnets should use value of 1, so please remove the test entry and start using 1

HoytRen:
Thank you very much for help. I will fix this problem later (maybe in a few days) because our community is voting for the relative issues. I may cancel the RP if the vote fails.

prusnak:
Ok open when you are ready

PR #1540, opened 2023-04-27
add Elys Network testnet bech32 prefix `elys` to slip-0173.md

ollieottersync:
Elys Network is currently in testnet, registering the `elys` prefix here prior adding to the cosmos chain registry. Our github is  https://github.com/elys-network

email: team@elys.network
twitter: @elys_network

### DIFF
--- a/slip-0173.md
+++ b/slip-0173.md
@@ -83,6 +83,7 @@ These are the registered human-readable parts for usage in Bech32 encoding of wi
 | Dyson Protocol           | `dys`         |          |             |
 | Echelon                  | `echelon`     |          |             |
 | e-Money                  | `emoney`      |          |             |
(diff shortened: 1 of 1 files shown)


cosmic-vagabond (approved):
LGTM

prusnak:
if you register `elys` for testnet, you won't be able to use `elys` for the mainnet, because we need prefixes to be different

is that what you want?

ollieottersync:
We want to use `elys` for mainnet. We are using it for testnet as well. Is it best that we don't register at all? Or should I register it as a mainnet prefix instead now?

prusnak:
Only mainnet

ollieottersync:
thank you will update to change to mainnet

PR #1546, opened 2023-05-08
Update slip-0044.md

GiMa-Maya:
Maya has the same derivation path as rune


prusnak:
We don't allow multiple coins to register the same number.

Are you suggesting we should remove entry 932 because it is not used.

prusnak:
> Not register Maya?

Yes

PR #1555, opened 2023-05-21
Create SLIP-0137: Like BIP 137 but it defines the sign/verify procedure for all address types.

ZenulAbidin:
This was a BIP I created to somewhat standardize the procedure for generating signed messages for post-P2PKH address types. You can see it [here on the mailing list](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-August/020826.html). It was unceremoniously rejected with the reason being that more effort should be directed to BIP 322.

None of the recent efforts on BIP 322 came into any fruition on the BIPs repo. 

Some companies such as Coinkite (the coldcard makers) are already using this BIP so it's best for all parties to officially publish this document in some capacity so that people don't have to rely on the source of the document being on my website, which can go down at any time.

Only superficial modifications such as changing "BIP" to "SLIP" have been made to the original BIP. I'm also open to deleting some of the optional modules if it means the core parts of the document get merged.

### DIFF
--- a/slip-0137.md
+++ b/slip-0137.md
@@ -1,0 +1,442 @@
+<pre>
+  SLIP: 0137
+  Layer: Applications 
+  Title: Procedure for signing and verifying messages
+  Author: Ali Sherief <ali@notatether.com> 
+  Status: Draft
+  Type: Informational
+  License: BSD-2-Clause
+  Created: 2022-08-04
+  Modified: 2023-05-21
+</pre>
+
+== Copyright ==
+
+This document is licensed under the 2-clause BSD license.
+
+== Abstract ==
+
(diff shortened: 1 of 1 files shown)


ZenulAbidin:
Some further questions:

- Should I drop Taproot address signing capability altogether, since signatures are not using ECDSA anyway?
- I have encountered some libraries that serialize messages to ASCII before signing with them. Since this is very bad for internationalization, should I make UTF8 encoding a requirement instead of optional?
-Maybe I could add a Recommendations section and move the parts about message text cleanup there, and state it should be a best practice for wallet software, instead of defining it as a format or extension of some sort?

ZenulAbidin:
Really guys? No comments?

Armss9936 (approved):

Armss9936 (approved):

debbiederrick7-cloud (approved):

PR #1567, opened 2023-06-13
Add Arthera

artheraone:
Add the Arthera network coin


artheraone:
https://docs.arthera.net/

PR #1578, opened 2023-07-19
Add GGXT

xcthulhu:


xcthulhu:
@prusnak We intend for our network to be incentivized like Kusama... is that acceptable?

PR #1580, opened 2023-08-03
codex32 hrps for slip-0173

roconnor-blockstream:
Codex32 applications uses the same sort of hrp/bech32 format that BIP-0173 uses.

The use of codex32 for master seeds, using the `ms` prefix, is described in BIP-0093.

CLN v23.08 is using the `cl` prefix for its codex32 HSM secret recovery format.

### DIFF
--- a/slip-0173.md
+++ b/slip-0173.md
@@ -233,6 +233,18 @@ other entries use Bech32.
 |       | `uivk` (m)                 | `uivktest` (m)             | `uivkregtest` (m)             |
 |       | `uview` (m)                | `uviewtest` (m)            | `uviewregtest` (m)            |
 
+## Uses of codex32
+
+The codex32 format is used to store master secret data. It features an extended
+checksum versus the one used in Bech32 in order to support enhanced error
+correction. Codex32 uses the same notion of a human-readable part and the same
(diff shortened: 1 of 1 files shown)


prusnak:
SLIP-0173 lists prefixes for applications that use BIP-0173 (Bech32) or BIP-0350 (Bech32m).
Since Codex32 doesn't use neither of these, I feel it does not belong int that document.

roconnor-blockstream:
IMHO, having a collection of used prefixes is a valuable resource. 

If someone where to start using the `ms`  prefix for their coin's bech32(m) address after consulting with SLIP-0173 to see if it is a free prefix, they would be liable to cause confusion with `ms` prefixed seed storage values of the form

    MS12NAMEA320ZYXWVUTSRQPNMLKJHGFEDCAXRPP870HKKQRM

and may cause people to run error correction on such a hypothetical coin's `ms`  bech32 addresses mistakenly thinking they are BIP-0093 master seeds.

That said, this is your SLIP repo, so it is ultimately your call.

prusnak:
> IMHO, having a collection of used prefixes is a valuable resource.

OK, reopened and merged.

PR #1591, opened 2023-08-30
Update slip-0044.md

nsonanh:
Add DHP (dHealth network) slip-044


prusnak:
We don't allow assigning one number to multiple coins. Choose another.

PR #1598, opened 2023-09-08
Add Enjin Coin

leonardocustodio:
Hello, I'm a developer at Enjin and we will be removing Efinity in favor of Enjin Blockchain. We would like to keep the 1155 id as that's a special number for us.


prusnak:
Provide a link to the Efinity website which says Efinity is going to be removed in favor of Enjin.

leonardocustodio:
You can check on from Efinity to Enjin section https://finance.yahoo.com/news/enjin-blockchain-finally-launching-13th-043300804.html
Also here is the referendum approved to swap all EFI tokens to ENJ https://efinity.io/governance/gp-2023-06-14

PR #1603, opened 2023-09-15
slip-0044: add networks by unique

CertainLach:
Unique https://unique.network/networks/#unique


prusnak:
Opal is a testnet, it should use number 1 as all other testnets.

I don't see anything about Sapphire on the website - is it a testnet?

CertainLach:
Yes, both opal and sapphire are testnets, sorry for missing that.

PR #1607, opened 2023-09-24
Update slip-0044.md

rastislavcore:
Updating 654 for Core Blockchain

### DIFF
--- a/slip-0044.md
+++ b/slip-0044.md
@@ -682,7 +682,7 @@ All these constants are used as hardened derivation.
 | 651        | 0x8000028b                    | SFX     | Safex Cash                        |
 | 652        | 0x8000028c                    | SFT     | Safex Token                       |
 | 653        | 0x8000028d                    | WSFX    | Wrapped Safex Cash                |
-| 654        | 0x8000028e                    | USDG    | US Digital Gold                   |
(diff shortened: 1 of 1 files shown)


prusnak:
we do not allow duplicates

use different number

rastislavcore:
Pavol, could you please clarify your approach for resolving conflicts related to the numbers? I couldn't locate any documentation that addresses this. From what I understand, we had the number first. See reference: https://github.com/core-coin/go-core/blob/a2349aa801bc5adb984781aff45e1f3b07b38ee4/accounts/hd.go#L31

Thank you

prusnak:
You are supposed to register the number before using it.

rastislavcore:
Thank you for your response. I appreciate the clarification on registering the number first. However, from my examination of the provided resources and documentation, it wasn't evident that this was a required step. Perhaps we might consider updating the documentation or providing more explicit guidelines for this to avoid any future confusion.

I still believe that our initial approach had its merits, as referenced in the earlier commit. Maybe we can discuss this further to reach a common understanding?

prusnak:
I am making no exceptions. Feel free to use the number assigned to another coin on your own, but I am not adding duplicate entries to the list.

rastislavcore:
Thank you for clarifying that the list follows the "first come, first served" rule.

PR #1617, opened 2023-10-13
slip-0044: Add Bitrock Blockchain

BitrockChain:


prusnak:
The value needs to be lower than 2147483648 to be a valid value for BIP-44

BitrockChain:
I make it lower :)

PR #1618, opened 2023-10-15
Add Bitrock Chain

BitrockChain:


prusnak:
Keep the list sorted according to num value and use the lowercase letters for hex values

BitrockChain:
Hi, updated it. 
Thank you.

prusnak:
Hex value is still not fully lower-case

BitrockChain:
Yes, i missed one character. now, everything is lower-case.

PR #1619, opened 2023-10-17
Add Stratos to slip-0173

alexstratos:


prusnak:
Testnet cannot have the same value as mainnet

alexstratos:
@prusnak thanks, updated.

PR #1636, opened 2023-11-22
Add Saga bech32 prefix to SLIPS

ashishchandr70:
Add Saga bech32 prefix to SLIPS


ashishchandr70:
Hi @prusnak I was still getting this ready for the rest of our team to review so it was still in draft mode. If there are changes to be made, would that be the same PR process or would it require something extra?

prusnak:
Yeah, send a new PR if neded.

PR #1645, opened 2023-12-14
Add Pundi X chain to slip-0173

zakir-code:


prusnak:
value for testnet can't be the same as for mainnet

PR #1650, opened 2023-12-24
Add CONET mainnet (224433) & testnet (224422)

CoNETProject:
Layer Minus Protocol:
https://doceng.conet.network/core-technology/point-to-point-encrypted-communication-protocol/layer-minus-protocol

DePIN and Privacy:
https://conet.network/

CONET Platform:
https://apps.conet.network/


prusnak:
You should use value 1 for all testnets. Remove the assignment on line 224422

PR #1656, opened 2024-01-11
Fix algorithm for master key derivation

karbyshev:
This fixes the broken indexation in the master key derivation algorithm. The error was introduced in commit 5b7a092


andrewkozlik (approved):
LGTM. Thanks @karbyshev!

PR #1665, opened 2024-01-23
Add Pylons prefix to slip-0173.md

romanv1812:


prusnak:
Testnet and mainnet prefixes cannot be the same.

romanv1812:
> Testnet and mainnet prefixes cannot be the same.

In this case, for now I have added a prefix only for the mainnet

Reply from a Pylons team member:
"I think it was the same. We decided to upgrade the test net in place to main net."

PR #1670, opened 2024-02-02
add hedge to slip-0173 file

crypto-niac:
This PR will add Hedge networks slips to slip-0173 file


prusnak:
testnet cannot be the same prefix as mainnet

crypto-niac:
> testnet cannot be the same prefix as mainnet

Updated.

cryptopunkstar:
Hey updated !

PR #1724, opened 2024-05-07
Add extendable backup flag in SLIP-0039

andrewkozlik:
@matejcik please review too.

### DIFF
--- a/slip-0039.md
+++ b/slip-0039.md
@@ -52,12 +52,14 @@ Notation        | Meaning
 *N<sub>i</sub>* | total number of members in group *i*, a positive integer, 1 &le; *N<sub>i</sub>* &le; 16
 *GT*            | group threshold, a positive integer, 1 &le; *GT* &le; *G*
 *T<sub>i</sub>* | member threshold for group *i*, a positive integer, 1 &le; *T<sub>i</sub>* &le; *N<sub>i</sub>*
-*id*            | random identifier, a 15-bit positive integer
-*MS*            | master secret, a string
+*id*            | random identifier, a 15-bit unsigned integer
+*ext*           | extendable backup flag
+*e*             | iteration exponent, a 4-bit unsigned integer
+*MS*            | master secret, an octet string
 *n*             | length of the master secret in bytes
-*EMS*           | encrypted master secret, a string
+*EMS*           | encrypted master secret, an octet string
 &#124;&#124;    | concatenation operator
-xor             | bit-wise exclusive-or of two strings
+xor             | bit-wise exclusive-or of two octet strings
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


slip-0039.md
```diff
@@ -254,12 +258,13 @@ The encrypted master secret is then `EMS = R || L`.
...
 ```
-F(i, R) = PBKDF2(PRF = HMAC-SHA256, Password = (i || passphrase), Salt = ("shamir" || id || R), iterations = 2500 << e, dkLen = n/2 bytes)
+F(i, R) = PBKDF2(PRF = HMAC-SHA256, Password = (i || passphrase), Salt = (salt_prefix || R), iterations = 2500 << e, dkLen = n/2 bytes)
 ```
 
 The value of *i* is encoded as one byte.
 
-The random identifier value *id* is encoded as two bytes in big-endian byte order.
+If *ext* = 1, then `salt_refix` is an empty string.
+If *ext* = 0, then `salt_refix = "shamir" || id`, where the random identifier value *id* is encoded as two bytes in big-endian byte order.
```
matejcik, line 267:
```suggestion
If *ext* = 1, then `salt_prefix` is an empty string.
If *ext* = 0, then `salt_prefix = "shamir" || id`, where the random identifier value *id* is encoded as two bytes in big-endian byte order.
```

slip-0039.md
```diff
@@ -403,6 +408,12 @@ Python wallets with SLIP39 support:
...
+
+    In newly created shares *ext* SHOULD be set to 1. However, since some users already have shares that use *ext* = 0, implementations MUST support master secret decryption using both *ext* values. If *ext* = 0, then implementations MUST NOT allow the intentional creation of another set of shares with the same *id* and *EMS*.
```
matejcik, line 415: > implementations MUST NOT allow the intentional creation 

if i were to nitpick, i'd say "SHOULD NOT" here, and even tighten the conditions to "same EMS + same distinguishing metadata" -- explicitly allowing one of the alternative behaviors, such as (a) extending a 1-of-1 backup or (b) bumping the group id for every newly created group

matejcik (commented):
As a general note, this is not 100% backwards compatible because the `ext` flag is not described as being part of the exponent in the earlier text. Perhaps a mention somewhere in the "Design rationale" section for it?

slip-0039.md
andrewkozlik: I suppose we could go with "SHOULD NOT". According to RFC 2119:
> there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.

- If a wallet takes adequate precautions, like allowing the upgrade only once in case (a) or guaranteeing strict monotonicity in case (b), while also guaranteeing in both cases that the upgrades cannot be done for recovered seeds, but only for seeds generated in the wallet, then it seems OK to do, see next point. Furthermore we'd need to clarify what exactly "same distinguishing metadata" means, which might be tricky. I wouldn't want to provide this sort of detailed guidance in the spec, so I'd stick with "EMS and id".
- However, what concerns me with being lenient is that if a user presents shares from mixed backups of the same seed, then wallets that are not aware of this possibility may have trouble guiding the user correctly in the recovery process. For example, in case (a) the user starts by entering a share from the newer multi-share backup and then enters the original 1-of-1 share. The wallet not expecting such a scenario will probably raise an exception, because the thresholds do not match, giving a cryptic error message to the user or possibly aborting the whole process, even though the situation is clearly resolvable. A similar problem may occur in case (b) with the group count. I wouldn't want to complicate this specification by outlining such possibilities and requiring wallets to take something like that into account.

andrewkozlik: > As a general note, this is not 100% backwards compatible because the ext flag is not described as being part of the exponent in the earlier text. Perhaps a mention somewhere in the "Design rationale" section for it?

Can you be more specific? You mean saying "note that the ext flag was added in a later revision, previously it was a bit of the iteration exponent", or also talk about the reasons why it was decided to revise SLIP 39 instead of creating a new one, or something else?

matejcik: > You mean saying "note that the ext flag was added in a later revision, previously it was a bit of the iteration exponent",

Just this, yes, in order to point out that you could encounter shares with ext flag set and `"shamir"` customization string, and there is a valid interpretation of that.

matejcik: so basically the question is which is the right thing to do here, informally speaking, "conformant wallets (SHOULD NOT/MUST NOT) produce data with high likelihood of confusing other conformant wallets", right?

I'd still lean towards "SHOULD NOT".
(maybe even drop the "and EMS" part?)
We could call out this problem by saying something like:
> implementations SHOULD NOT allow the intentional creation of another set of shares with the same *id* and *EMS*. Doing so would run the risk of the users trying to recover shares from incompatible groups, which is something that implementations are not required to handle.

andrewkozlik: [af7a2b0](https://github.com/satoshilabs/slips/pull/1724/commits/af7a2b03fe1eca57197c1d0c013e7177be06b522) should address all of the points discussed.

matejcik (approved):

PR #1725, opened 2024-05-07
Update slip-0044.md

lastmeta:
Evrmore, a fork of Ravencoin uses the same coin type value as Ravencoin.


prusnak:
We do not allow multiple entries per number. Choose another value for your coin.

PR #1733, opened 2024-05-27
add Neura to the SLIP-0173

memekas:


memekas:
@prusnak testnet and regtest prefixes have been deleted

PR #1737, opened 2024-05-31
feat: register nillion bech32 prefix on slip 173

glottologist:


prusnak:
You can't have the same prefix for testnet. Change it.

PR #1738, opened 2024-06-04
Add Bech32 for BlockX

ahmadSaeedGoda:
- Register the human-readable parts for usage in Bech32 for BlockX Network/Chain.


prusnak:
You can't use the same prefix for mainnet and testnet.

ahmadSaeedGoda:
@prusnak Thanks, excuse me that's not clear in the requirements. And it is the actual prefix for both networks indeed.
So if you kindly let me understand why, I will definitely go change it for the testnet for example then come back here to add the changed one.
Removed the testnet prefix for now to register the mainnet, Kindly find the last commit.
Thanks & Regards,

PR #1741, opened 2024-06-12
Add coin type for Qubic

couch42:


couch42:
Wanted to follow up on the status of this PR. What do you need to get this PR accepted?

Some additional context: we are having a working MetaMask Snaps integration but we need the coin type to be ready to publish. Also multiple other wallets ready.

You can find more informations about the project here: https://www.qubic.org

PR #1745, opened 2024-06-20
Add bech32 for HeliChain

HeatApp-Corp:


HeatApp-Corp:
#https://github.com/cosmos/chain-registry/pull/4757

PR #1775, opened 2024-08-10
Add Lumos coin type to SLIP-0044

gaokanxu:
Lumos is a block chain system written by Rust.


gaokanxu:
Lumos website https://www.uuhb.com。 We are developing a new type block chain system, more faster, cheaper than all existed system.

PR #1785, opened 2024-09-03
Register Integritee TEER in  slip-0044.md

brenzi:
Integritee is a parachain on Polkadot and Kusama. The native token for both networks is TEER


prusnak:
Remove the empty line you added

PR #1801, opened 2024-10-07
Add Whitechain network

whitechain-dev:


prusnak:
use the lowercase hex letters

PR #1813, opened 2024-11-02
Update slip-0044.md

daochild:
Remove record 122 

Reason:
https://www.icc-cpi.int/news/situation-ukraine-icc-judges-issue-arrest-warrants-against-vladimir-vladimirovich-putin-and?fbclid=IwAR0nGUmKFEJFyz1yebHaDKXOBtR3vQjinjCpn_e2GNnN9AdwGmhm2lE8YxU


prusnak:
We are not removing coins from the list to keep the space occupied so no other coins can reuse the number and cause confusion.

daochild:
Accept changes, commit doesn't break your rules.

prusnak:
> Accept changes, commit doesn't break your rules.

It does. I explained that in the post above.

PR #1814, opened 2024-11-03
multi: add missing Blackcoin info

BlackcoinDev:
I noticed Blackcoin is in SLIP-0044 but not yet in other SLIPs.


prusnak:
Added to slip-0132. No need to add it to other slips, since the coin is not very relevant.

PR #1815, opened 2024-11-04
Update `slip-0044.md`

breathx:
Add Vara (VARA) coin type for BIP-0044


prusnak:
All testnets should use value `1` as specified by BIP-44.

Please remove the testnet entry.

PR #1824, opened 2024-11-30
BIP-44: Add MST blockchain coin type

dev-svikrant406:


prusnak:
testnets should use the value 1 as per bip-44

please remove the testnet entry

dev-svikrant406:
@prusnak Thanks for your feedback :), I updated the PR as you suggested.

PR #1828, opened 2024-12-01
Update slip-0044.md to add slip44 for sei

Armingodiz:
I have added this based on coinId from https://github.com/trustwallet/wallet-core


prusnak:
Something is not right. Either the decimal value or the hex value are wrong.

Armingodiz:
The problem was hex value, which is fixed now

PR #1835, opened 2024-12-09
Changed references to BIP32 with references to SLIP-0010

hbs:


slip-0013.md
andrewkozlik, line 1: 1. Please replace `bip32 node path` with `HD node path` in the worked example below.
2. Please replace BIP-32 with SLIP-10 in the References section.

slip-0017.md
andrewkozlik, line 1: 1. Please replace BIP-32 with SLIP-10 in the References section at the end of the SLIP.
2. If you have the time, it would be nice to have a worked example here like we have in SLIP-0013.

andrewkozlik (changes requested):

hbs:
Please Take Another Look

andrewkozlik (approved):

PR #1848, opened 2025-01-08
update slip-0039.md implementations listed to reflect state of things

Huge:
the Bluewallet could likely be found in some spec. commit, but would that be worthy?
should we include https://github.com/iancoleman/shamir39/blob/master/src/js/shamir39.js ?


Armss9936 (approved):

Armss9936 (approved):

prusnak:
I don't like the change

Also shamir39 by ian coleman is not slip39 based

PR #1861, opened 2025-02-08
add slipid44 for DYDX

murdererbrain:
This PR adds support for the dYdX network to SLIP-0044 .

PR #1862, opened 2025-02-09
Add DYDX network SLIP-44

ramezanzade1:
Please add DYDX network Slip-44
Thanks


prusnak:
closing in favor of https://github.com/satoshilabs/slips/pull/1861 which is more sane

PR #1881, opened 2025-04-17
Add text details memo to SLIP-0024

andrewkozlik:
Related to github.com/trezor/trezor-firmware/pull/4960

### DIFF
--- a/slip-0024.md
+++ b/slip-0024.md
@@ -189,11 +189,13 @@ TODO
 
 ## Verifying address ownership
 
-This section is non-normative. One of the requirements in processing a refund memo or a coin purchase memo is that the customer's wallet must verify that the specified address is controlled by the wallet. The most common way of verifying address ownership by a wallet is to provide the BIP-0032 derivation path leading from the wallet's seed to the address. In cases where the address uses a script such as multisig, specifying the address derivation can be more complicated.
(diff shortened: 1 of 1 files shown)


slip-0024.md
```diff
@@ -188,12 +210,15 @@ k = Key(m/"SLIP-0024"/"Address MAC key")
...
+mac = HMAC-SHA256(key = k, msg = addressInfo)
 ```
 
-where `coinType || address` is the concatenation of the following fields:
+where `addressInfo` is the concatenation of the following fields:
 
 * *coinType* (4 bytes): 32-bit encoding of the SLIP-0044 coin type of the address in little-endian byte order.
+* The BIP-0032 path to derive the address from the master node, which consists of:
+    * *n* (CompactSize integer): the path length.
+    * *i*<sub>1</sub> || *i*<sub>2</sub> || ... || *i*<sub>*n*</sub> (variable length): concatenation of the 32-bit encodings of the child indices, each in little-endian byte order.
```
matejcik, line 221: i don't fully understand the role of adding this to the `addressInfo`

if i understand the address verification mechanism, the wallet may not even have the derivation path available at the time of processing the payment request? the host would have to tell it, along with the address authentication code.

what sort of distinguishing feature are we adding here?

andrewkozlik: > the host would have to tell it, along with the address authentication code.

Correct, and the host should have this information readily available.

I think this feature originates from a discussion between @tomasklim and me. The reason is that we may want to show the user the path to the address when they click on details in Trezor.

matejcik (approved):

PR #1890, opened 2025-05-16
slip-0010.md : add links to BIP-0032

gap-editor:


prusnak:
the BIP is linked in the `Motivation` paragraph, the change seems unnecessary

PR #1894, opened 2025-05-30
add MarsDoge

yerikokay:
add MarsDoge


prusnak:
0x05abd is not 23251 so which one do you want to register?

prusnak:
no response

PR #1902, opened 2025-06-11
SLIP-44: Add RGB smart contracts

dr-orlovsky:
Based on RGB-44 standard <https://github.com/RGB-WG/RFC/blob/master/RGB-44.md>


dr-orlovsky:
Yes, there is: RGB run on top of bitcoin, using its UTXOs with bitcoins, assigning additional assets to them client-side. 

Thus, it is important that non-RGB-aware bitcoin wallets won't be able to see or spent such outputs (otherwise RGB assets will be lost).

Dedicated testnet cointype is required in order to test this behavior in testnet.

PR #1905, opened 2025-06-19
Add: baby to slip-44

ryokubozono:
Added BABY which is native token of babylon chain.
https://docs.babylonlabs.io/guides/overview/babylon_genesis/baby_tokenomics/#overview


angelaryals (commented):
.

PR #1907, opened 2025-06-24
slip-0044: Add RAILGUN_

emilbayes:
This is for the derivation paths used by RAILGUN_ software:

https://github.com/Railgun-Community/engine/blob/3ae608337095046d926aabc3cb0eda2f1507cc8d/src/key-derivation/wallet-node.ts#L17-L20

https://railgun.ch


prusnak:
1984 is already registered by another coin and we do not allow duplicates

choose another number

PR #1908, opened 2025-06-30
Update slip-0039.md

super-e:
Add Slip39DotNet to C# implementations


prusnak:
I prefer to not recommend the code written by AI for now

PR #1909, opened 2025-07-07
slip-0021: Specify encoding of integer labels

M1nd3r:
Specify the encoding of integer labels: ASCII encoding as a decimal integer in the fewest possible number of characters.


andrewkozlik (commented):
IIRC, we started using we started using ASCII-encoded as a decimal integer in the fewest possible number of characters.

PR #1970, opened 2026-02-07
Remove name from entry 0x8000024b (incognito privacy)

0xmatt11:
Remove token name PRV


prusnak:
why?

also use `| ---     | reserved                          |` instead of empty

prusnak:
no response

PR #2002, opened 2026-04-07
rename: Coreum to TX in slip-0044 and slip-0173

TxCorpi0x:
## Description

The main change is renaming the entry previously labeled as "Coreum" to "TX" while retaining the same symbol and encoding identifiers.

**Specification updates:**

* Updated the name for coin type `0x800003de` from "Coreum" to "TX" in `slip-0044.md` to reflect the correct project name.
* Changed the human-readable part registration from "Coreum" to "TX" for the `core` prefix in `slip-0173.md`, maintaining the same Bech32 encoding.


TxCorpi0x:
@prusnak, Hi, May you please check this PR, the change is in Organizational level.
Old: [Coreum](https://github.com/CoreumFoundation/coreum) // https://coreum.com
New  [TX](https://github.com/tokenize-x/tx-chain) // https://tx.org

PR #2017, opened 2026-05-07
SLIP-0027: Variable-length keys for cryptographic schemes

andrewkozlik:


slip-0027.md
```diff
@@ -0,0 +1,178 @@
...
+ || ...
+ || HMAC-SHA512(key = Key(N), msg = retry || block_count - 1)
+```
+
+and the input parameters satisfy:
+
+* *retry* is a single byte, 0 ≤ *retry* ≤ 255.
+* *length* is the desired output length in bytes, an integer with 1 ≤ *length* ≤ 16384.
+* The block index is encoded as a single byte. The two-byte HMAC message *retry* || *block* is the byte *retry* followed by the byte *block*.
+
```
M1nd3r, line 62: I think we should highlight the importance of not using the Key(N) of a node as a keying material and using the same node as an input for the Expand function.

```suggestion

**Warning:** Never use `Key(N)` as output keying material and the same node *N* as input to `Expand`.


```

PR #2055, opened 2026-08-26
slip-0044: add MONEU (coin type 8328)

natusor:
Adds MONEU to the registry.

The wallet implements BIP-44 with derivation path m/44'/8328'/0'/0/index

Source: https://github.com/natusor/MONEU


slip-0044.md
GriegoG: What is this about and who is this let me know l like see what b this is about thank you

PR #2058, opened 2026-08-31
Add AithorChain (AIT) coin type 657384

Malikfaisaly:


prusnak:
keep the list sorted, do not add extra newline

PR #2061, opened 2026-09-10
Add coin type 1031 for Orange (OAG)

Kazuhiro-Tokumoto:
Add coin type 1031 for Orange (OAG)

Orange is a RandomX proof-of-work UTXO chain (Rust, public repo):
https://github.com/Kazuhiro-Tokumoto/Orange

BIP-39 / BIP-32 / BIP-44 are implemented in its wallet:
  crates/oag-wallet/src/bip39.rs   BIP-39 mnemonic (12 words)
  crates/oag-wallet/src/bip32.rs   BIP-32 derivation
  crates/oag-wallet/src/seed.rs    BIP-44 path

Both are verified against the official BIP-39 and BIP-32 test vectors.

Intended path: m/44'/1031'/0'/0/i

Note: the wallet currently refuses to derive mainnet keys, because no coin
type is assigned yet. Testnet uses the reserved coin type 1 and works today.
We deliberately do not ship a provisional number, since changing it later
would silently change the keys derived from existing backups. 1031 will be
pinned once assigned.

1031 is currently unused (gap between 1030 and 1032).
Neither "OAG" nor "Orange" appears in the current list.


Wyrelade:
1031 already requested earlier by me. PR up.

Kazuhiro-Tokumoto:
Thank you very much. I'll change the number

